PR #100, opened 2022-11-14
feat: Remove social account for ticketCredential [DEV-1921]

DaevMithran:


ankurdotb:
Task linked: [DEV-1921 Remove credential guard for IIW credential](https://app.clickup.com/t/6600954/DEV-1921)

src/api/controllers/credentials.ts
```diff
@@ -61,8 +61,59 @@ export class Credentials {
 		})
 	}
 
-	async issue_credentials(request: Request, user: GenericAuthUser, provider: string, subjectId?: string): Promise<Response> {
-		provider = provider.toLowerCase()
+	async issue_person_credential(user: GenericAuthUser, provider: string, subjectId?: string) {
```
jay-dee7, line 64: @DaevMithran add the return type on the method signature

src/api/controllers/credentials.ts
```diff
@@ -61,8 +61,59 @@ export class Credentials {
...
+			id: subjectId,
+			type: undefined
+		}
+
+		const credential = {
+			'@context': VC_CONTEXT.concat(VC_PERSON_CONTEXT),
+			type: ['Person', VC_TYPE],
+			issuanceDate: new Date().toISOString(),
+			credentialSubject: credential_subject,
+			'WebPage': [
```
jay-dee7, line 75: @DaevMithran did we decide to go with this format of `WebPage`? AFAIK, platforms like discord don't have a notion of `name` and `URL`. This is why we had the switch/case here before, to make the `WebPage` property adjust to the provider

src/api/controllers/credentials.ts
DaevMithran: Sure got it, I'll change that, or is it better to remove the URL itself instead of hardcoding these?

src/api/controllers/credentials.ts
DaevMithran: Sure got it, I'll change that, or is it better to remove the URL itself instead of hardcoding these?

jay-dee7 (approved):
LGTM 💯 @rosspower11 I'm merging this one in. If we get any feedback we'll merge that in too

PR #111, opened 2022-11-24
feat: Support JSONLD [DEV-1931]

DaevMithran:
Facing this issue in issuing a jsonLD credential
```jsonld.InvalidUrl: Dereferencing a URL did not result in a valid JSON-LD object. Possible causes are an inaccessible URL perhaps due to a same-origin policy (ensure the server uses CORS if you are using client-side JavaScript), too many redirects, a non-JSON response, or more than one HTTP Link Header was provided for a remote context.```

Pending
Fix the build.yml github workflow

### CODE REVIEW

**Comment by ankurdotb** [comment]
Task linked: [DEV-1931 Build full support for JSON-LD credentials into credential-service & setup veramo to user this](https://app.clickup.com/t/6600954/DEV-1931)

File: Dockerfile [hunk 1]
```diff
@@ -1,8 +1,8 @@
 ###############################################################
-###         STAGE 1: Build credential-service app           ###
+###                  Build credential-service app           ###
```
ankurdotb, line 2: Can you please use a two-stage build like in upstream? Helps optimise Docker image size.

package.json
```diff
@@ -8,8 +8,8 @@
 	"author": "Cheqd Foundation Limited (https://github.com/cheqd)",
 	"repository": "https://github.com/cheqd/credential-service.git",
 	"scripts": {
-		"build": "webpack",
-		"start": "miniflare dist/worker.js --global-async-io --global-timers --global-random",
+		"build": "tsc",
```
ankurdotb, line 11: If you're not using TypeScript, then we can change this but it seems kinda weird to do a TypeScript compile and then use the index.js source file instead (since we're copying from `dist/src` in Dockerfile rather than the compiled output.

package.json
```diff
@@ -31,18 +31,19 @@
 		"@veramo/key-manager": "^4.2.0",
 		"@veramo/kms-local": "^4.2.0",
 		"did-resolver": "^4.0.1",
+		"dotenv": "^16.0.3",
```
ankurdotb, line 34: Why do you need this? Environment variables can generally just be read from environment right?

ankurdotb (changes requested):

Tweeddalex:
Closing as JSON-LD issuance is now supported in Veramo

PR #277, opened 2023-06-27
feat: Allow unauthenticated users to use some functionality [DEV-2851]

anikitinDSR:


ankurdotb:
Task linked: [DEV-2851 Credential Service: Unauthenticated mode](https://app.clickup.com/t/6600954/DEV-2851)

DaevMithran:
These changes are gonna result in only a `Customer Not found error` as the createAgent function here expects the customerId https://github.com/cheqd/credential-service/blob/main/src/services/identity/postgres.ts#L218

Three possible solutions discussed with @anikitinDSR 

- Modify the initAgent function in PostgressIdentity class (set enableCredentials to true, modify types, in public functions use this.agent instead of createAgent)
- Create a separate class for Unauthenticated, still have to handle the type issues as ICheqd type can't be assigned without providers
- Update the localIdentity class, **Best approach**, if the mnemonic is not provided in .env https://github.com/cheqd/credential-service/blob/main/src/services/identity/local.ts#L43, instead of returning error skip creating providers

PR #286, opened 2023-07-06
feat: Add Default role assigning while creating and account or signing in [DEV-2905]

anikitinDSR:


ankurdotb:
Task linked: [DEV-2905 Automatically assign Testnet User role to users and top-up the user's Cosmos/cheqd account](https://app.clickup.com/t/6600954/DEV-2905)

src/app.ts
```diff
@@ -72,22 +73,25 @@ class App {
...
     )
-    this.express.use(Authentication.handleError)
-    this.express.use(Authentication.accessControl)
+    this.express.use(auth.handleError)
+    this.express.use(async (req, res, next) => await auth.accessControl(req, res, next))
   }
 
   private routes() {
     const app = this.express
+    const jsonMiddleware = express.json({ limit: '50mb' })
```
DaevMithran, line 82: This is not correct.

parsing middlewares are used according to the content type of request body. Need not separate them out and specify in each api

DaevMithran (changes requested):

src/app.ts
```diff
@@ -72,12 +74,14 @@ class App {
...
     )
-    this.express.use(Authentication.handleError)
-    this.express.use(Authentication.accessControl)
+    this.express.use(auth.handleError)
+    this.express.use(async (req, res, next) => await auth.accessControl(req, res, next))
   }
 
   private routes() {
     const app = this.express
+    const rawMiddleware = express.raw({ type: 'application/octet-stream' })
```
DaevMithran, line 83: Please move this middeware to the middleware() function. we need not declare a variable and use it in the route.

DaevMithran (changes requested):

DaevMithran (dismissed):

src/middleware/auth/logto.ts
```diff
@@ -50,6 +49,135 @@ export class LogToHelper {
...
+
+        
+    } catch (err) {
+        return {
+            error: `getRolesForUser ${err}`,
+            status: 500,
+            data: {}
+        }
+    }
+  }
```
ankurdotb, line 144: Honestly this is a more complicated way of doing things. Ask for the role ID, then it's a simple role ID GET. https://docs.logto.io/api/#tag/Roles/paths/~1api~1roles~1:id/get

Try to match a name is an inexact science and gets into the realm of getting case of the string right or wrong etc.

src/middleware/auth/logto.ts
```diff
@@ -50,6 +49,135 @@ export class LogToHelper {
...
```
ankurdotb, line 99: Seems unnecessarily complex to use the Roles API and then check all users, rather than going under the Users API, for a specific user, and then using the GET and POST endpoints: https://docs.logto.io/api/#tag/Users/paths/~1api~1users~1:userId~1roles/post

ankurdotb (changes requested):

src/middleware/auth/base_auth.ts
```diff
@@ -40,9 +40,7 @@ export abstract class AbstractAuthHandler implements IAuthResourceHandler
         this.token = '' as string
         this.scopes = undefined
         this.customer_id = '' as string
-        if (process.env.ENABLE_AUTHENTICATION === 'true') {
-            this.logToHelper = new LogToHelper()
-        }
+        this.logToHelper = new LogToHelper()
```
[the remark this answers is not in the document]
anikitinDSR: No, for now all the Class makes sense only if authentication is enabled. It checks on the top level

ankurdotb (dismissed):

PR #295, opened 2023-07-11
fix: Fix issue with broken webhook being used for setting default user role in LogTo [DEV-2962]

abdulla-ashurov:
- [x] Fixed issue with broken webhook.


ankurdotb:
Task linked: [DEV-2962 Fix issue with broken webhook being used for setting default user role in logto](https://app.clickup.com/t/6600954/DEV-2962)

PR #304, opened 2023-07-25
fix: Fix verifyCredential && resolve function

DaevMithran:
- allows application/x-www-form-urlencoded
- remove decodeCredential from jwt, as it's not required. 
- set's credentialStatus policy to false, if verifyStatus is false

### DIFF
--- a/src/services/identity/agent.ts
+++ b/src/services/identity/agent.ts
@@ -354,7 +354,7 @@ export class Veramo {
  }
 
  async resolve(didUrl: string) {
-    const result = await fetch(process.env.RESOLVER_URL || resolverUrl + didUrl, {
+    const result = await fetch((process.env.RESOLVER_URL || resolverUrl) + didUrl, {
         headers: { 'Content-Type': 'application/did+ld+json' },
     })
     const ddo = (await result.json())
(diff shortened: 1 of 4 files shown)


src/services/identity/agent.ts
```diff
@@ -240,22 +239,23 @@ export class Veramo {
...
             }
         } as ICheqdVerifyCredentialWithStatusList2021Args)
     } else {
         result = await agent.verifyCredential({
             credential, 
             ...verificationOptions,
-            fetchRemoteContexts: verificationOptions.fetchRemoteContexts || decodedCredential.proof.jws
+            policies: {
+                ...verificationOptions.policies,
+                credentialStatus: false,
```
anikitinDSR, line 257: why did we add here `credentialStatus: false`? Does it mean that we want to make such field not mandatory?

DaevMithran: credentialStatus true is handled in the above condition which call's did-provider-cheqd cheqdVerifyCredential

the native verifyCredential doesn't support our status. if verifyStatus is set to false in request, it move to this condition

anikitinDSR (approved):

abdulla-ashurov (approved):

PR #306, opened 2023-07-26
feat: Account bootstrapping [DEV-3051]

anikitinDSR:
- Add CredentialAccount creating if it's not created yet
- Add default role assigning if the list of roles is empty
- Add token populating if there are absent or not enough even for DID creation


src/controllers/customer.ts
```diff
@@ -106,4 +119,78 @@ export class AccountController {
...
+                if (resp.status !== StatusCodes.OK) {
+                    return response.status(StatusCodes.BAD_GATEWAY).json({
+                        error: resp.error})
+                }
+            }
+        }
+        return response.status(StatusCodes.OK).json({})
+    }
+
+    public static async checkBalance(address: string, network = NetworkType.Testnet) {
```
DaevMithran, line 190: @anikitinDSR It's better to move this as a util function to our sdk. Even the getCosmosAccount. As these functions can be resused and can be managed easily when there are breaking changes in cosmjs

src/services/customer.ts
```diff
@@ -22,8 +22,14 @@ export class CustomerService {
             throw new Error('Customer exists')
         }
         const kid = (await new Identity(customerId).agent.createKey('Secp256k1', customerId)).kid
+        const address = getCosmosAccount(kid)
```
DaevMithran, line 25: getCosmosAccount should be move to sdk util as well

DaevMithran (changes requested):
Looks good, requested to move cosmjs functions entirely to sdk utils

abdulla-ashurov (changes requested):

DaevMithran (approved):

abdulla-ashurov (approved):

PR #307, opened 2023-07-27
fix: Cleanups [DEV-3050]

anikitinDSR:
Get API guarding back:
    - removed /credential-status/check
    - removed /credential-status/search
    - removed /did resolve the DID
    - removed /presentation/verify


ankurdotb:
Task linked: [DEV-3050 Credential Service: Cleanup and minor fixes](https://app.clickup.com/t/6600954/DEV-3050)

src/services/identity/index.ts
```diff
@@ -3,12 +3,14 @@ import { PostgresIdentity } from './postgres.js'
...
 dotenv.config()
 
 export class Identity {
     private agent: LocalIdentity | PostgresIdentity
 
     static instance = new Identity().agent
+    static unauthorized = new Unauthorized()
 
     constructor() {
         if (process.env.ENABLE_EXTERNAL_DB === 'true') {
```
DaevMithran, line 16: @anikitinDSR 

Instead of this approach, can we add an optional argument to the constructor as `agentId?: string`

```
if (process.env.ENABLE_EXTERNAL_DB === 'true' ) {  
this.agent = agentId ? new PostgresIdentity(agentId) : Identity.unauthorized
 }
```
This will be useful once we integrate payments, as only authorized users can do payments. Hardcoding `Identity.unauthorized ` will be a problem 

Note we pass the agentId argument within PostgresIdentity

DaevMithran (approved):

PR #317, opened 2023-08-03
fix: Move to using scopes correctly from LogTo [DEV-3090]

anikitinDSR:
- Move to asking scope directly from logto
- refactorings and cleanups


ankurdotb:
Task linked: [DEV-3090 Change current authorisation process from using accessToken to asking LogTo with M2M](https://app.clickup.com/t/6600954/DEV-3090)

abdulla-ashurov (changes requested):

src/middleware/auth/base-auth.ts
```diff
@@ -59,24 +59,25 @@ export abstract class AbstractAuthHandler implements IAuthResourceHandler
...
+                return this.returnError(400, "Unauthorized error: Seems like you are not authenticated. Please follow the authentication process using 'LogIn' button")
             }
-            const resourceAPI = AbstractAuthHandler.buildResourceAPIUrl(request)
-            if (!resourceAPI) {
-                return this.returnError(500, `Internal error. Issue with building resource API for the path ${request.path}`)
+            // Tries to get customerId from the logTo user structure
+            if (request.user && request.user.claims) {
+                this.customerId = request.user.claims.sub
+            } else {
+                return this.returnError(500, "Unauthorized error: Seems like authentication process was corrupted and there are problems with getting customerId")
```
DaevMithran, line 70: if we are reusing, would be good to store them as error message constants

DaevMithran (changes requested):

src/middleware/auth/base-auth.ts
anikitinDSR: It's a unique error message :) Or do you mean that we need to define all the error messages as constants?

abdulla-ashurov (dismissed):
A good job! 👍

src/middleware/auth/base-auth.ts
```diff
@@ -43,7 +30,7 @@ export abstract class AbstractAuthHandler implements IAuthResourceHandler
         this.namespace = '' as Namespaces
         this.token = '' as string
         this.scopes = []
-        this.customer_id = '' as string
+        this.customerId = '' as string
```
ankurdotb, line 33: Good catch! This is a great example of making naming conventions consistent (and following TypeScript style guidelines)

src/middleware/auth/base-auth.ts
ankurdotb: I like the thinking here @DaevMithran but storing the error message text somewhere else could make the error return less obvious when reading code (since I'd then have to look up the constants file). A better idea is to use a convention/format when writing error messages, rather than picking a random text (but keep it within the context of where the error is sent from).

So I don't think you need to move this to a constants file, but you guys do need to get consistent with what HTTP status codes you use. :) Since this is true in a couple of other places, I'll leave it as an overall comment.

src/middleware/auth/base-auth.ts
```diff
@@ -59,24 +46,25 @@ export abstract class AbstractAuthHandler implements IAuthResourceHandler
...
+                return this.returnError(400, "Unauthorized error: Seems like you are not authenticated. Please follow the authentication process using 'LogIn' button")
             }
-            const resourceAPI = AbstractAuthHandler.buildResourceAPIUrl(request)
-            if (!resourceAPI) {
-                return this.returnError(500, `Internal error. Issue with building resource API for the path ${request.path}`)
+            // Tries to get customerId from the logTo user structure
+            if (request.user && request.user.claims) {
+                this.customerId = request.user.claims.sub
+            } else {
+                return this.returnError(500, "Unauthorized error: Seems like authentication process was corrupted and there are problems with getting customerId")
```
ankurdotb, line 57: 500 is the wrong error code for this type: either this is not an "Unauthorised" case, or the error code needs to change.

src/middleware/auth/base-auth.ts
```diff
@@ -59,24 +46,25 @@ export abstract class AbstractAuthHandler implements IAuthResourceHandler
...
-            // Tries to get token from the request and other preps
-            const _setup = this.setupAuthToken(request)
-
-            // If token is not provided - try to proceed without it
-            // If token is provided - verify it and get the scopes
-            if (_setup.status !== 200) {
-                return _setup 
+            // If the user is not authenticated - return error
+            if (!request.user.isAuthenticated) {
+                return this.returnError(400, "Unauthorized error: Seems like you are not authenticated. Please follow the authentication process using 'LogIn' button")
```
ankurdotb, line 51: Incorrect status code

src/middleware/auth/logto.ts
```diff
@@ -49,83 +49,96 @@ export class LogToHelper {
     return this.allResourceWithNames
   }
 
-  public async setDefaultRoleForUser(userId: string): Promise<ILogToErrorResponse | void> {
+  public async setDefaultRoleForUser(userId: string): Promise<ILogToErrorResponse> {
     const roles = await this.getRolesForUser(userId)
-    if (roles && roles.status === 200) {
+    if (roles.status === 200) {
```
ankurdotb, line 54: Would cross-check with @abdulla-ashurov since he worked on this, but is the status code still 200 when a user has no roles? (This is the current logic as far as I follow, checks that the response actually has roles in it, not just whether status code is 200)

src/middleware/auth/logto.ts
```diff
@@ -239,60 +228,52 @@ export class LogToHelper {
...
+        const data = await response.json();
+        this.m2mToken = data.access_token
+        return this.returnOk({})
     } catch (err) {
-        return {
-            status: 500,
-            error: `setM2MToken: ${err}`,
-            data: {}
-        }
+        return this.returnError(500, `setM2MToken: ${err}`)
```
ankurdotb, line 239: Will this be thrown to an end user ever? If yes, check whether there's a risk of the error message revealing too much about the server. If it only logs within say DO, this is fine, if it gets bubbled up to the user might be a bad idea.

ankurdotb (changes requested):
This is overall a good PR and shows progress, nicely done @anikitinDSR!

I want to make a general point though that I see in this PR, but also in what @DaevMithran sometimes does in his PRs: you're often using 5xx status codes and 4xx status codes pretty much interchangeably. Best practice is NOT to do that. 4xx codes and 5xx codes have very different meanings. Here are some tangible examples:

1. In some places, we're returning an error 400. 400 stands for "Bad Request", i.e., it specifically means the **client application** sent a malformed request that the server can't understand. The error message next it stated it was "Unauthorised". If the case you're handling is about auth errors, you should be using 401 Unauthorised _if and only if_ it relates to an auth failure between client and server. In our case, this would mean if the user is not logged in. Note the subtle difference between 401 and 403 in the spec: 401 is to be used when the client is not authenticated at all, but 403 is used when the client _is_ authenticated, but doesn't have the right permissions. See [HTTP 4xx codes on MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses).
2. Note that 4xx errors are indications about **mistakes/errors the client has made** when talking to the server. If you have pure server side failures, not specific to a user/client, then the correct ranges [typically are in 5xx for internal server failures](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#server_error_responses). Normally, your status codes would fall into something specific in 2xx, 3xx, or 4xx range, so a 5xx error is typically a default/catchall scenario, if and only if there wasn't a more specific error to return. See below. Typically, a 5xx range is not used when responding to the client since it's often a severe, unhandled failure/exception. (You should still "catch" it with a 5xx properly; a bit like in Python where you probably don't want a programme to publish an unhandled error/stack trace which might reveal sensitive details.)

While not relevant here, I've seen this sometimes in creds-app too where 301, 302, 308 etc are used very interchangeably. While they all mean redirect, semantically, they are very different things.

Take a look at [LogTo's APIs](https://docs.logto.io/api/). You'll see:
1. 400 if the request body was wrong
2. 401 if the request wasn't authenticated (i.e., are they logged in or not). You'll see in MDN they clarify that although the term is '401 Unauthorized', what it means in practice is not authenticated. (Authentication and authorisation are different things.)
3. 403 if they _are_ logged in, but don't have the right permissions.
4. 500 as a catch-all: This is a much more severe error, like a server crash. When a proxy like Nginx or Cloudflare hit a 5xx code, it will typically assume the server has had a catastrophic error and crashed while processing the request. You wouldn't, commonly, be communicating this to the client. (In fact, Cloudflare will override any 5xx codes and serve its own page quite often, rather than one provided by you.)

DaevMithran:
@ankurdotb we are already following this pattern for error statuses

src/middleware/auth/base-auth.ts
anikitinDSR: So, here it would be a real 500 error (maybe text should be changed) cause here I assume that LogTo middleware added `user` property for the `request` object. Such state is not under the client control

src/middleware/auth/logto.ts
anikitinDSR: Status == 200 here means that `this.getRolesForUser(userId)`  method finished successful. I made such internal structure on purpose cause here we could pass the same structure as a response to user with the same code and use the `error` field as the reason of error.
Also, for the first bootstrap thing we need to check is there any roles assigned to the user as a sanity check and if it's empty - assign TestNet role.

DaevMithran:
LGTM

PR #325, opened 2023-08-08
feat: Change from `did/:did` to `did/resolve/:didUrl` endpoint and support query parameters for DID endpoint in Credential Service [DEV-3097]

abdulla-ashurov:
- [x] Change DID endpoint from `did/:did` to `did/resolve/:did`.
- [x] Update swagger OpenAPI.
- [x] Support query and fragment options for DID endpoint.
- [x] Add new GET resource endpoints:
  - [x] `/resource/list/:did` - for getting the list of resources.
  - [x] `/resource/dereferencing/:did/:resourceId` - for getting the specific resource.


ankurdotb:
Task linked: [DEV-3097 Change from `did/:did` to `did/search/:did` endpoint and support query parameters for DID endpoint in Credential Service](https://app.clickup.com/t/6600954/DEV-3097)

src/app.ts
```diff
@@ -115,15 +117,17 @@ class App {
     app.post(`/did/update`, IssuerController.updateValidator, new IssuerController().updateDid)
     app.post(`/did/deactivate/:did`, IssuerController.deactivateValidator, new IssuerController().deactivateDid)
     app.get(`/did/list`, new IssuerController().getDids)
-    app.get(`/did/:did`, new IssuerController().getDid)
+    app.get(`/did/resolve/:didUrl`, new IssuerController().getDid)
 
     // Resource API
     app.post(`/resource/create/:did`, IssuerController.resourceValidator, new IssuerController().createResource)
+    app.get(`/resource/list/:did`, new IssuerController().getResourceList)
+    app.get(`/resource/dereferencing/:did/:resourceId`, new IssuerController().getResource)
```
DaevMithran, line 125: @abdulla-ashurov can we remove /dereferencing from this api path. It's make it complicated to understand

DaevMithran (changes requested):

src/controllers/issuer.ts
```diff
@@ -287,36 +287,37 @@ export class IssuerController {
       if (request.body.didDocument) {
         updatedDocument = request.body.didDocument
       } else if (did && (service || verificationMethod || authentication)) {
-        const resolvedResult = await new Identity(response.locals.customerId).agent.resolveDid(did)
-        if(!resolvedResult?.didDocument || resolvedResult.didDocumentMetadata.deactivated) {
+        const body = await (await fetch(`${process.env.RESOLVER_URL}/${did}`)).json()
+        const resolvedResult: DIDResolutionResult = JSON.parse(body)
+        if (!resolvedResult?.didDocument || resolvedResult.didDocumentMetadata.deactivated) {
           return response.status(StatusCodes.BAD_REQUEST).send({
```
anikitinDSR, line 293: I don't think that here the error needs to be BAD_REQUEST. Mostly BAD_GATEWAY cause such request is not handled by CredentialService side but by did-resolver.

src/services/identity/IIdentity.ts
```diff
@@ -1,42 +1,126 @@
...
+	BulkRevocationResult, BulkSuspensionResult, BulkUnsuspensionResult, CreateStatusList2021Result,
+	RevocationResult, StatusCheckResult, SuspensionResult, UnsuspensionResult,
+} from '@cheqd/did-provider-cheqd/build/types/agent/ICheqd'
+import type {
+	BroadCastStatusListOptions, CheckStatusListOptions, CreateStatusListOptions, CredentialRequest,
+	StatusOptions, UpdateStatusListOptions, VeramoAgent, VerificationOptions,
+} from '../../types/types'
 
 export interface IIdentity {
-  agent?: TAgent<any>
```
anikitinDSR, line 18: Why did you remove `agent` and `privateStore` from interface?

src/services/identity/default.ts
```diff
@@ -0,0 +1,40 @@
...
+	checkStatusList2021(did: string, statusOptions: CheckStatusListOptions, agentId?: string): Promise<StatusCheckResult> {
+		return Veramo.instance.checkStatusList2021(this.initAgent(), did, statusOptions)
+	}
+
+	searchStatusList2021(did: string, statusListName: string, statusPurpose: 'revocation' | 'suspension', agentId?: string): Promise<any> {
+		return Veramo.instance.searchStatusList2021(this.initAgent(), did, statusListName, statusPurpose)
+	}
+}
+
+async function fetchResponseBody(url: string): Promise<[string, string]> {
```
anikitinDSR, line 36: Let's move it to the `helpers`

src/services/identity/local.ts
```diff
@@ -30,20 +29,18 @@ const {
   ISSUER_DID
 } = process.env
 
-export class LocalIdentity implements IIdentity {
-  agent?: VeramoAgent
-  privateStore?: AbstractPrivateKeyStore
-
+export class LocalIdentity extends DefaultIdentity {
   constructor() {
```
anikitinDSR, line 33: Am I right that such constructor is the same as for all child classes also? If yes - let's move it to the default class

anikitinDSR (changes requested):
Hey! Great work.
Let's have a look at review comments and also please take into account that adding new routes even they are should not be guarded we need to add rule for it.

src/services/identity/index.ts
```diff
@@ -23,12 +22,12 @@ export class Identity {
     }
 
     public setupIdentityStrategy(agentId?: string) {
-        if (process.env.ENABLE_EXTERNAL_DB === 'true') {
-            if (agentId) {
+        if (agentId) {
```
anikitinDSR, line 25: Let's revert it back

src/controllers/issuer.ts

src/services/identity/default.ts
abdulla-ashurov: Done! ✅

src/services/identity/index.ts
abdulla-ashurov: Done! ✅

src/services/identity/local.ts
abdulla-ashurov: We cannot move it to default constructor of `DefaultIdentity` class because when `DefaultIdentity` calls own constructor then Virtual Table is not fully complete, and it means: if I put constructor implementation to `DefaultIdentity` class default constructor, then inside `DefaultIdentity` constructor will call unimplemented `DefaultAgent::initAgent()` method.

abdulla-ashurov: Done ✅

src/services/identity/IIdentity.ts
abdulla-ashurov: If I move it `IIdentity` interface that I need to redefine it `AbstractIdentity` class.

src/app.ts
abdulla-ashurov: We cannot rename from `resource/dereferencing/:did/:resourceId` to `resource/:did/:resourceId` because if we remove the `dereferencing` from endpoint, we end up with 2 endpoints that will be blocked for each other:
1. `/resource/:did/:resourceId`: it handles everything which comes after `resource/`
2. `/resource/list/:did` - this endpoint never works because all requests from this endpoint will be handled by the first endpoint.

PR #341, opened 2023-08-15
fix: Add `serviceId` and `serviceType` sections for DID Document creation for swagger [DEV-3125]

abdulla-ashurov:
- [x] Make default DID Document null for simplifying DID creation.
- [x] Add `serviceId` and `serviceType` fields for `application/x-www-form-urlencoded`


ankurdotb:
Task linked: [DEV-3125 Add serviceId and serviceType sections for DID Document creation for swagger](https://app.clickup.com/t/6600954/DEV-3125)

src/controllers/issuer.ts
```diff
@@ -255,8 +257,8 @@ export class IssuerController {
 				if (serviceEndpoint) {
 					didDocument.service = [
 						{
-							id: `${didDocument.id}#service-1`,
-							type: 'service-1',
+							id: didDocument.id+"#"+serviceId,
```
anikitinDSR, line 260: Let's do it in formatting style, like:
`${didDocuemnt.id}#${serviceId}`

src/controllers/issuer.ts
```diff
@@ -13,7 +13,7 @@ import { generateDidDoc, validateSpecCompliantPayload } from '../helpers/helpers
 export class IssuerController {
 	public static createValidator = [
 		check('didDocument')
-			.optional()
+			.optional({nullable: true})
```
anikitinDSR, line 16: Why do we need it here? As I know we already allowed to pass empty DIDDocument, did we?

anikitinDSR (changes requested):

src/controllers/issuer.ts

src/controllers/issuer.ts

PR #344, opened 2023-08-16
fix: Check API responses are provided correctly in Swagger document [DEV-3133]

abdulla-ashurov:
- [x] Make correct APIs response formats
- [x] Sort swagger OpenAPI tags.


ankurdotb:
Task linked: [DEV-3133 Check API responses are provided correctly in Swagger document](https://app.clickup.com/t/6600954/DEV-3133)

src/controllers/issuer.ts
```diff
@@ -388,7 +388,8 @@ export class IssuerController {
 	 *         content:
 	 *           application/json:
 	 *             schema:
-	 *               $ref: '#/components/schemas/DidResult'
+	 *               type: boolean
+	 *               example: true
```
ankurdotb, line 392: Is this actually the result that it returns, i.e., a boolean value of true/false without being actually nested in a JSON key-value? If the answer is yes, we should:

1. On deactivation, the ideal response would be to send back the DID document (like in a normal GET for a DID) since this will have the property `deactivated` set to true. This might require a change in the code logic.
2. If the above is done, then I believe the schema reference will go back to what it was originally.

src/controllers/revocation.ts
```diff
@@ -255,18 +255,7 @@ export class RevocationController {
...
-	 *                 properties:
-	 *                   statusListName:
-	 *                     type: string
-	 *                   statusListVersion:
-	 *                     type: string
-	 *                   statusListId:
-	 *                     type: string
-	 *                   statusListNextVersion:
-	 *                     type: string
+	 *               $ref: '#/components/schemas/CredentialStatusSearchResponse'
```
ankurdotb, line 258: I do believe this response is accurate. Could you make the result for a status list search return an array of objects similar to the response that `/create` does? That one is obviously just returning one, this would return the same fields but multiple ones in an array. (Skipping the `created: true` property)

src/controllers/revocation.ts
```diff
@@ -342,7 +331,7 @@ export class RevocationController {
 	 *         content:
 	 *           application/json:
 	 *             schema:
-	 *               $ref: '#/components/schemas/CredentialStatusResult'
+	 *               $ref: '#/components/schemas/CredentialStatusUpdateResponse'
```
ankurdotb, line 334: Similar story here, could we change this so that an `/update` returns _very_ similar to what `/create` does? Except the `created: true` will change to `updated: true`

src/types/swagger-types.ts
```diff
@@ -786,6 +793,38 @@
...
+ *           type: object
+ *         id:
+ *           type: string
+ *           example: 398cee0a-efac-4643-9f4c-74c48c72a14b
+ *         name:
+ *           type: string
+ *           example: ResourceName
+ *         resourceType:
+ *           type: string
+ *           example: TextDocument
```
ankurdotb, line 812: If this is what the code is returning, can we rationalise it to return a body similar to what `linkedResourceMetadata` for one resource object is? No sense in doing a reduced subset that's been selected out, also means we could reuse the schema perhaps.

src/types/swagger-types.ts
```diff
@@ -786,6 +793,38 @@
...
+ *           statusListId:
+ *             type: string
+ *           statusListName:
+ *             type: string
+ *           statusListVersion:
+ *             type: string
+ *           statusListNextVersion:
+ *             type: string
+ *           statusPurpose:
+ *             type: string
```
ankurdotb, line 827: Already commented on this in `revocation.ts`

ankurdotb (changes requested):

src/controllers/issuer.ts

src/controllers/revocation.ts
abdulla-ashurov: Done! ✅

src/types/swagger-types.ts
abdulla-ashurov: Done! ✅

src/types/swagger-types.ts
abdulla-ashurov: Done! ✅

src/controllers/revocation.ts
abdulla-ashurov: Done! ✅

src/types/swagger-types.ts
abdulla-ashurov: Done! ✅

ankurdotb (changes requested):
Getting there, but still not quite there yet.

- Currently, there are four different schemas for Credential Status. This part is fine, since the outermost wrapper for each response _is_ different
   - `/create` uses `CredentialStatusResult`
   - `/update` uses `CredentialStatusUpdateRequest`
   - Skipping publish because we're removing it in #350 
   - `/search` uses `CredentialStatusSearchResponse`
- Each of the schemas then defines its own body (i.e., they repeat the body section). But the only thing that's _actually_ changing is:
   - `/create` has the string `created: true` on top
   - `/update` has `updated: true` on top
   - That's it! The rest of the body _should_ be the same (They might not in the current APIs)
- Another point of inconsistency (unsure whether this comes from Swagger, or the actual response:
  - For `/create` and `/update` the response is structured like this:
  
  ```json
  {
  created: true,
  resource: { 
     StatusList2021: ...,
     metadata
  },
  resourceMetadata
  ```
  - For `/search` this changes to
  
  ```json
  {
  resource: ...,
  resourceMetadata
  }
  ```
  
The difference is that in search, for some reason, at least in example and in Swagger, the `resource` key no longer has `StatusList2021` or `metadata` nested underneath. Why? This should theoretically be a response from resolution.

Here's a better way to structure credential-status APIs (as an example):
1. Have a **single** definition in the types file of Credential Status resource, which includes resource + resource metadata.
2. In `create`, add the `created: true` boolean property, and then reuse this definition above
3. In `update`, add the `updated: true` property, and reuse same definition as before
4. Search can directly use the definition under 1.
5. The benefit here is that there's a single, harmonised definition of what a credential status resource with its metadata looks like, and only differing in the response on the unique bits.

If the actual response is different than structure above, then this is a good sanity check that the response is probably wrong, and should be fixed to match the schema above.

Our current approach makes the schema too verbose, since we're copy pasting the schema components multiple times, which also makes it very hard to get a grip on whether the issue is the response is incorrect, or whether the docs are incorrect.

PR #345, opened 2023-08-17
fix: Add DID deactivation check for credential issuing [DEV-3136]

abdulla-ashurov:
- [x] Add parameter `allowDeactivatedDid` for allowing/permitting verify credential with `deactivatedDID` in `/credential/verify` endpoint. 
- [x] Add parameter `allowDeactivatedDid` for allowing/permitting verify credential with `deactivatedDid` in `/presentation/verify` endpoint.
- [x] Block issue credential based on a deactivated DID in `/credential/issue` endpoint.


ankurdotb:
Task linked: [DEV-3136 Add DID deactivation check for credential issuing](https://app.clickup.com/t/6600954/DEV-3136)

Tweeddalex:
@anikitinDSR please let me know when this has been reviewed by you :)

src/controllers/credentials.ts
```diff
@@ -165,7 +180,35 @@ export class CredentialController {
...
+				// If it's not a JWT - just skip it
+				if (!(e instanceof InvalidTokenError)) {
+					throw e;
+				}
+			}
+		}
+
+		if (!allowDeactivatedDid) {
+			const result = await new IdentityServiceStrategySetup(response.locals.customerId).agent.resolveDid(issuerDid);
+			if (!result?.didDocument || result.didDocumentMetadata.deactivated) {
```
anikitinDSR, line 205: I think the check here should be more wider. Like what if the `result` or  `didDocumentMetadata` are empty?

anikitinDSR: in such cases you may provide the error from `resolveDid`

src/controllers/credentials.ts
```diff
@@ -423,7 +472,35 @@ export class CredentialController {
...
+		const allowDeactivatedDid = request.query.allowDeactivatedDid === "true";
+
+		let issuerDid = "";
+		let decoded: any;
+
+		if (presentation.issuer?.id) {
+			issuerDid = presentation.issuer.id;
+		} else {
+			try {
+				decoded = jwt_decode(presentation);
```
anikitinDSR, line 485: The same comment, let's move it to the helper function

src/controllers/credentials.ts
```diff
@@ -101,6 +103,13 @@ export class CredentialController {
 			request.body['@context'] = [request.body['@context']];
 		}
 
+		const resolvedResult = await new IdentityServiceStrategySetup(response.locals.customerId).agent.resolveDid(request.body.issuerDid);
+		if (!resolvedResult?.didDocument || resolvedResult.didDocumentMetadata.deactivated) {
+			return response.status(StatusCodes.BAD_REQUEST).send({
+				error: `${request.body.issuerDid} is either Deactivated or Not found`,
```
ankurdotb, line 109: I agree with @anikitinDSR below actually. Collapsing the `deactivated` and `not found` conditions results in a less helpful message to the user (which case was it?). I would split these to:

1. DID _does_ exist, but it's deactivated. Don't allow issuing a credential in this scenario, no overrides. This kind of check might be better to carry out in did-provider-cheqd. Does Veramo not automatically do this behaviour, i.e., does the built-in `issue` method skip checking whether a DID is deactivated or not? @Eengineer1 
2. A separate condition to test whether a DID was not found. Separating this allow for more helpful error messages. This is available upstream because `resolveDid` has `notFound` property for `DIDResolutionMetadata`

src/controllers/credentials.ts
```diff
@@ -165,7 +180,35 @@ export class CredentialController {
...
+		const verifyStatus = request.query.verifyStatus === 'true';
+		const allowDeactivatedDid = request.query.allowDeactivatedDid === "true";
+
+		let issuerDid = "";
+		let decoded: any;
+
+		if (credential.issuer?.id) {
+			issuerDid = credential.issuer.id;
+		} else {
+			try {
```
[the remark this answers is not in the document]
ankurdotb: `allowDeactivatedDid` is needed in both credential as well as presentation verify. (Presentation verify might need to resolve multiple DIDs, since the credentials contained might be from different issuers.)

So instead of thinking of this as a helper function specific to credential service, should this not be in did-provider-cheqd? Instead of doing a `resolveDid` in credential service? It could be passed under `verificationOptions` or am i missing something?

src/controllers/credentials.ts
ankurdotb: Agreed and good catch @anikitinDSR! I added similar comments above.

ankurdotb (changes requested):
Besides the other comments, the query

src/controllers/credentials.ts
abdulla-ashurov: Done! ✅

src/controllers/credentials.ts
abdulla-ashurov: Moved to the helper.ts. ✅

src/controllers/credentials.ts
abdulla-ashurov: Done second request! ✅

PR #351, opened 2023-08-23
fix: Validate did access on operation

DaevMithran:


src/services/identity/postgres.ts
```diff
@@ -232,36 +240,69 @@ export class PostgresIdentityService extends DefaultIdentityService {
...
 		const agent = await this.createAgent(agentId)
+        if (!await CustomerService.instance.find(agentId, { did })) {
+            throw new Error(`${did} not found in wallet`)
+        }
 		return await Veramo.instance.updateStatusList2021(agent, did, statusOptions, publish)
 	}
 
 	async checkStatusList2021(did: string, statusOptions: CheckStatusListOptions, agentId: string): Promise<StatusCheckResult> {
 		const agent = await this.createAgent(agentId)
+        if (!await CustomerService.instance.find(agentId, { did })) {
```
anikitinDSR, line 259: I think for checking the statusList we don't need to enable such guarding cause it's public info. Am I right?

src/services/identity/postgres.ts
```diff
@@ -232,36 +240,69 @@ export class PostgresIdentityService extends DefaultIdentityService {
...
+            throw new Error(`${did} not found in wallet`)
+        }
 		return await Veramo.instance.broadcastStatusList2021(agent, did, resourceOptions, statusOptions)
 	}
 
 	async revokeCredentials(credentials: VerifiableCredential | VerifiableCredential[], publish: boolean, agentId: string) {
 		const agent = await this.createAgent(agentId)
+        for(var credential in credentials) {
```
anikitinDSR, line 275: If it's a common functionality let's make it as a function, maybe helper

DaevMithran: made it as a function within the same class as it's not gonna be used in others

anikitinDSR (approved):

anikitinDSR (approved):

PR #364, opened 2023-09-04
test: Credential-Service: Add integration tests for unauthorized scenario [DEV-3141]

abdulla-ashurov:
- [x] Add positive integration tests for 6 APIs:
  - [x] /did/search
  - [x] /resource/search
  - [x] /credential/verify
  - [x] /credential-status/search
  - [x] /credential-status/check
  - [x] /presentation/verify (**Added TODO comment**)
- [x] Add negative integration tests for 6 APIs:
  - [x] /did/search
  - [x] /resource/search
  - [x] /credential/verify
  - [x] /credential-status/search
  - [x] /credential-status/check
  - [x] /presentation/verify (**Added TODO comment**)


ankurdotb:
Task linked: [DEV-3141 Credential-Service: Add integration tests for unauthorized scenario](https://app.clickup.com/t/6600954/DEV-3141)

tests/constants.ts
```diff
@@ -13,3 +44,95 @@ export const DEFAULT_SUBJECT_DID = 'did:key:z6MkhaXgBZDvotDkL5257faiztiGiC2QtKLG
...
+		type: "JwtProof2020"
+	},
+	type: [
+		"VerifiableCredential",
+		"Person"
+	]
+};
+
+export const INVALID_JWT_TOKEN = "invalid_jwt_token";
+export const INVALID_CREDENTIAL = {
```
anikitinDSR, line 113: Let's not to keep such large values as a constant. It's hard to maintain them in the future. Also, please take it into account that I introduced builders for general requests and let's keep the same logic for responses also.

tests/credential-status/check.no-auth.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+    NOT_EXISTENT_STATUS_LIST_NAME
+} from '../constants';
+import * as fs from 'fs';
+import { test, expect } from '@playwright/test';
+import { StatusCodes } from 'http-status-codes';
+
+test('[Positive] It can check an unencrypted status-list with an existent body and statusPurpose=revocation parameter', async ({ request }) => {
+    const response = await request.post('/credential-status/check?statusPurpose=revocation', {
+        data: {
+            did: TESTNET_DID,
```
anikitinDSR, line 15: From my point of view I think it's better to use builders here and in the similar places for other tests.
builder here could be a function, which accepts all the important variables and returns the expected `Request` object.

tests/credential-status/check.no-auth.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+            index: 10,
+            statusListName: "cheqd-employee-credentials",
+        },
+        headers: { 'Content-Type': CONTENT_TYPE.APPLICATION_JSON }
+    });
+    expect(response.status()).toBe(StatusCodes.OK);
+
+    const body = await response.json();
+    const expected = JSON.parse(
```
anikitinDSR, line 24: And also here. As @ankurdotb already mentioned, we need to check not all the fields but only specific list. I think it's better to use here some specific `expect` functions, like `expect_credential_status_list_unencrypted` which could check only significant fields

tests/credential-status/search.no-auth.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+    const response = await request.get(
+        '/credential-status/search?' +
+        `did=${TESTNET_DID}&` +
+        'statusPurpose=revocation&' +
+        'statusListName=cheqd-employee-credentials'
+    );
+    expect(response.status()).toBe(StatusCodes.OK);
+
+    const body = await response.json();
+    const expected = JSON.parse(
```
anikitinDSR, line 22: The same situation here as before. Cause we are using the real testnet DIDs in case if this DID be changed we will get false-negative scenario here.

anikitinDSR (changes requested):

tests/credential-status/check.no-auth.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+//     });
+//     expect(response.status()).toBe(StatusCodes.OK);
+
+//     const body = await response.json();
+//     const expected = JSON.parse(
+//         fs.readFileSync(`${PAYLOADS_PATH.CREDENTIAL_STATUS}/check-status-suspension.json`, 'utf-8')
+//     );
+
+//     expect(body).toStrictEqual(expected);
+// });
```
ankurdotb, line 89: Encrypted status lists cannot be checked without an account

tests/credential-status/check.no-auth.spec.ts
ankurdotb: Agreed, as said before, don't expect `strictEquals` to work since the body may change. You don't have to test for everything, but you _can _ test for things like whether the DID matches, the name matches, the type matches etc.

tests/constants.ts
```diff
@@ -1,6 +1,37 @@
...
+	APPLICATION_JSON = "application/json",
+	APPLICATION_DID_LD_JSON = "application/did+ld+json"
+};
+
+export const DID_METHOD = "cheqd";
+
 export const DEFAULT_MAINNET_DID = 'did:cheqd:mainnet:7c950b5d-dbbb-4a12-9d79-6b553ca0c271';
 export const DEFAULT_TESTNET_DID = 'did:cheqd:testnet:7bf81a20-633c-4cc7-bc4a-5a45801005e0';
 
+export const TESTNET_DID = "did:cheqd:testnet:d4a13003-0bc5-4608-b23a-54ea90fe9f90";
```
ankurdotb, line 21: Do you need both `DEFAULT_TESTNET_DID` and `TESTNET_DID`? Probably not.

tests/credential/verify.no-auth.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+        data: { credential: VALID_JWT_TOKEN },
+        headers: { 'Content-Type': CONTENT_TYPE.APPLICATION_JSON }
+    });
+    expect(response.status()).toBe(StatusCodes.OK);
+
+    const body = await response.json();
+    const expected = JSON.parse(fs.readFileSync(`${PAYLOADS_PATH.CREDENTIAL}/verify-jwt.json`, 'utf-8'));
+
+    expect(body).toStrictEqual(expected);
+});
```
ankurdotb, line 26: Might not always work, test key elements of the response instead

tests/did/search.no-auth.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+            created: TESTNET_DID_CREATED_TIME
+        },
+        contentMetadata: {}
+    };
+
+    expect(body.dereferencingMetadata.contentType).toBe(expected.dereferencingMetadata.contentType);
+    expect(body.dereferencingMetadata.did).toStrictEqual(expected.dereferencingMetadata.did);
+    expect(body.contentStream.created).toBe(expected.contentStream.created);
+    expect(body.contentMetadata).toStrictEqual(expected.contentMetadata);
+});
```
ankurdotb, line 63: This is very fragile and can break easily

tests/payloads/credential-status/check-status-revocation.json
```diff
@@ -0,0 +1,3 @@
+{
+    "revoked": false
+}
```
ankurdotb, line 3: Honestly, you don't need a file for this (parse and check for the value)

tests/payloads/credential-status/check-status-suspension.json
```diff
@@ -0,0 +1,3 @@
+{
+    "suspended": false
+}
```
ankurdotb, line 3: Don't need a file for this

tests/payloads/credential-status/search-list-revocation.json
```diff
@@ -0,0 +1,28 @@
...
+        "resourceName": "cheqd-employee-credentials",
+        "resourceType": "StatusList2021Revocation",
+        "mediaType": "application/json",
+        "resourceVersion": "2023-09-07T11:01:49.443Z",
+        "created": "2023-09-07T11:01:47Z",
+        "checksum": "32ffe033eb9d6c5c0d0c84dca06a9fc0b07980e1dfc3e1bc733dc6366ca781f0",
+        "previousVersionId": "dafbe3d2-85a8-4fba-b726-a21cd4b5e381",
+        "nextVersionId": null
+    }
+}
```
ankurdotb, line 28: Don't do a file with `strictEquals` on the whole file. If you want, you can keep this file and then compare specific key-values against the response.

tests/presentation/verify.no-auth.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { test } from '@playwright/test';
```
ankurdotb, line 1: Tests missing here, this is definitely in scope of something we can check right now.

tests/resource/search.no-auth.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+import { test, expect } from '@playwright/test';
+import { StatusCodes } from 'http-status-codes';
+
+test('[Positive] It can search resource with an existent DID and resourceId', async ({ request }) => {
+    const response = await request.get(`/resource/search/${TESTNET_DID}?resourceId=${TESTNET_DID_RESOURCE_ID}`);
+    expect(response.status()).toBe(StatusCodes.OK);
+
+    const body = (await response.body()).toString();
+
+    expect(body).toEqual(TESTNET_RESOURCE);
```
ankurdotb, line 21: Test for a known image or JSON instead, that way you can also check whether MIME headers are passed correctly.

ankurdotb (changes requested):

PR #367, opened 2023-09-05
test: Playwright integration test for RBAC [DEV-3164]

anikitinDSR:


ankurdotb:
Task linked: [DEV-3164 Write integration RBAC tests reusing tests written by Abdulla](https://app.clickup.com/t/6600954/DEV-3164)

tests/fixtures.ts
```diff
@@ -0,0 +1,110 @@
...
+
+   DIDRepository: [
+    async ({ browser }, use) => {
+        expect(browser.isConnected()).toBeTruthy();
+        const didFixture = new DIDFixture();
+        await didFixture.createDid();
+        await use(didFixture);
+    }, 
+    { scope: 'worker'}]
+})
```
ankurdotb, line 110: I get the idea behind the DIDFixture a bit more, but I'm not sure if it's used in the intended way. A good example of a fixture would be for example collating repeatable elements and then using them in tests using extends. So a good example of a fixture would be:

1. In global setup, do stuff like login etc.
2. Create fixtures such as:
  1. You can successfully GET an account (this proves you're logged in)
  2. Account has balance
  3. A given DID exists
4. For DID tests, call the fixtures 1 and 2 above before attempting a DID create (since this is async, DID creation will not be kicked off until those conditions are satisfied.)
5. Call fixture 3 before attempting DID update
6. Call fixtures 1 and 3 before doing anything under Credentials API

You get the idea. It's a sanity check before you proceed with other test actions.

I wouldn't make DID _creation_ a fixture because it's a test on its own. _Checking_ it exists is a good use of a fixture though, because it ensures the test failure isn't because a race condition happened, and something like credential status create was attempted before the DID for it existed.

tests/setup.ts
```diff
@@ -0,0 +1,34 @@
...
+    await page.getByPlaceholder('Password').fill(process.env.TEST_USER_PASSWORD);
+    await page.getByRole('button', { name: 'Continue' }).click();
+    await page.waitForURL(`${process.env.APPLICATION_BASE_URL}/swagger/`, {
+        waitUntil: 'domcontentloaded',
+        timeout: 60000,
+    });
+
+    // End of authentication steps.
+	await page.context().storageState({ path: storageState });
+}
```
ankurdotb, line 32: Move logging in to [auth.setup.ts](https://playwright.dev/docs/auth) instead. Check whether you need to explicitly call it from globalSetup, or if it's automatically invoked.

Where we'll _really_ need auth.setup.ts if/when we want to run one set of tests as a testnet user, and the same set of tests as mainnet user. Which basically means parallel workers, logged in as different test users. https://playwright.dev/docs/auth#multiple-signed-in-roles

tests/setup.ts
```diff
@@ -0,0 +1,34 @@
...
+async function globalSetup(config: FullConfig) {
+    const { storageState } = config.projects[0].use;
+
+    const chromiumBrowser = await chromium.launch();
+    await setupPage(chromiumBrowser, storageState as string)
+}
+
+async function setupPage(browser: Browser, storageState: string) {
+    const page = await browser.newPage();
+    await page.goto(`${process.env.APPLICATION_BASE_URL}/logto/sign-in`);
```
ankurdotb, line 16: Isn't it better to go to Swagger page first, and click the Login button? That would be more accurate to what users do.

tests/setup.ts
```diff
@@ -0,0 +1,34 @@
...
+	});
+
+    await page.getByPlaceholder('Email / Username').fill(process.env.TEST_USER_EMAIL);
+    await page.getByRole('button', { name: 'Sign in' }).click();
+    await page.getByPlaceholder('Password').fill(process.env.TEST_USER_PASSWORD);
+    await page.getByRole('button', { name: 'Continue' }).click();
+    await page.waitForURL(`${process.env.APPLICATION_BASE_URL}/swagger/`, {
+        waitUntil: 'domcontentloaded',
+        timeout: 60000,
+    });
```
ankurdotb, line 28: Assumption here is that login was successful after redirect, but that may not always be the case. A better test is that Swagger page was loaded, and a button called "Logout" exists (don't click it).

tests/setup.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    // End of authentication steps.
+	await page.context().storageState({ path: storageState });
+}
+
+export default globalSetup;
```
ankurdotb, line 34: Besides logging in, what might be other things we could do in global setup? Some examples I can think of are:

1. Once you're logged in, check if GET account returns customer ID and a cheqd address. (The button might have changed after logging in, but that's no guarantee that you're _actually_ logged in until you can get an account response.) Once we expand the DB table for customers, you could probably check for other elements.

I have a few other ideas, but tbh they require changes to the Account & Key APIs **so we should skip these for now** but other things could have been:
1. IMO, the GET account API could/should also return the account balance in CHEQ, which would allow us in global setup to test whether there's sufficient balance.
2. If you actually click the Account create API, it throws an error because there's already an associated account. The original intention of course was to create new Cosmos/cheqd accounts under the same Customer ID.
3. Key create is also kinda meaningless, since we always generate new keys for every DID create request (for example) and don't allow people to specify a key alias to use. If we actually supported specifying an existing key, then perhaps one of the global setup steps would have been to create a new key before the execution of further tests (and then using that key in a DID create).

ankurdotb (changes requested):

tests/authorized/rbac/credential-status.spec.ts
```diff
@@ -0,0 +1,63 @@
+import { test } from '../../fixtures';
+import { expect } from '@playwright/test';
+import { StatusCodes } from 'http-status-codes';
+import { DEFAULT_DOES_NOT_HAVE_PERMISSIONS } from '../../constants';
+import * as fs from 'fs';
+
+const PAYLOADS_BASE_PATH="./tests/authorized/payloads/rbac/credential-status";
+
```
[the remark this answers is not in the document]
anikitinDSR: With return code I also check the error in response. In case if it was rejected because of RBAC the message should say about it

tests/authorized/rbac/credential-status.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+test('Create encrypted StatusList2021 for user with testnet role but network is mainnet', async ({ request }) => {
+    const response = await request.post(
+        `/credential-status/create/encrypted`, 
+        {
+            data: JSON.parse(fs.readFileSync(`${PAYLOADS_BASE_PATH}/credential-status-create-encrypted-without-permissions.json`, 'utf-8')),
+            headers: {
+                "Content-Type": "application/json",
+            },
+        });
```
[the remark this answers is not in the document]
anikitinDSR: Am I right that we still wanna to check the JSON also? I did it in that way as we do it in cheqd-node. To make it more clear to define what the JSON is sending through such test.

tests/fixtures.ts
```diff
@@ -0,0 +1,110 @@
...
+            process.env.APPLICATION_BASE_URL + path, 
+            {
+                body: JSON.stringify(data),
+                headers: headers,
+                method: 'PATCH',
+            }
+        );
+        return response;
+    }
+}
```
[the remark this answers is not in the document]
anikitinDSR: request is test-scoped fixture but I need here worker-scoped. For fixture issue I could potentially comment it out cause such fixtures could help Abdulla with authorised tests.

tests/fixtures.ts
anikitinDSR: I think that DID creation fixture could be useful at least for checking the general list of steps from user for issuing the credential. Like CreateDID->CreateStatusList->IssueCredential->Revoke/suspend. So for all of this steps we could use DID fixture which automatically checks the DID creation and could be reused for all of other checks.

PR #440, opened 2023-12-06
fix: Json & Cors validation

DaevMithran:
This PR fixes the below issues

- As the expected property is JSON, use `isJSON()` instead of `isObject()`

- The current code expects CORS_ALLOWED_ORIGINS to be an array, as it's an env variable it should be a comma separated string


ankurdotb:
Task linked: [DEV-3400 Add `makeFeePayment` parameter for `/presentation/verify`](https://app.clickup.com/t/6600954/DEV-3400)

PR #442, opened 2023-12-11
test: Add credential API tests [DEV-3238]

DaevMithran:


ankurdotb:
Task linked: [DEV-3238 Add tests to the /credential/issue API](https://app.clickup.com/t/6600954/DEV-3238)

anikitinDSR (changes requested):
Hey! Thanks for your work. I think it worse to make some extra checks.
1. For issued credentials let's check the info that can be verified at this point, like:
- issuer is as expected,
- credentialSubject is the same
- credentialStatus field is placed and has expected type
- type includes `VerifiableCredential ` type
2. Also, Maybe it worse to double check that all the changes are made for credential statusList objects also. Let's check also, that revocation/suspension/reinstantion also reflect to statusList changes.

anikitinDSR (approved):

PR #454, opened 2023-12-29
fix: Make all the response bodies in predictable format [DEV-3494]

anikitinDSR:
- Cover responses with corresponding types
- Cover request bodies, params and queries with corresponding types
- Add absent validations + small improvements


ankurdotb:
Task linked: [DEV-3494 Make response bodies in predictable format](https://app.clickup.com/t/6600954/DEV-3494)

src/controllers/credentials.ts
```diff
@@ -391,22 +429,25 @@ export class CredentialController {
...
+
 		// Get strategy e.g. postgres or local
 		const identityServiceStrategySetup = new IdentityServiceStrategySetup(response.locals.customer.customerId);
 
 		try {
 			const result = await identityServiceStrategySetup.agent.suspendCredentials(
-				request.body.credential,
-				publish,
+				credential,
+				publish as boolean,
```
anikitinDSR, line 448: @Eengineer1 Could you please take a look here as well. I'm not sure is it right decision in terms of TypeScript ideology. Validator should transform this parameter into `boolean` type, cause there is a call `toBoolean` but request.query is still a structure of strings. So, I check it in debug session and `publish` variable has explicit type `boolean` and it seems like ok, but anyway.

Eengineer1: Well tracked down and yes, I do prefer assertions, especially if there's no type narrowing involved, in which case this syntax is well suited here.

src/controllers/did.ts
```diff
@@ -286,32 +307,27 @@ export class DIDController {
...
-		const { did, service, verificationMethod, authentication } = request.body as {
-			did: string;
-			service: Service[];
-			verificationMethod: VerificationMethod[];
-			authentication: string[];
-		};
-		let updatedDocument: DIDDocument;
+		const { did, service, verificationMethod, authentication } = request.body as UpdateDidRequestBody;
+		// Get the didDocument from the request if it's placed there
+		let updatedDocument: DIDDocument | undefined = request.body.didDocument ;
```
Eengineer1, line 318: I'd refactor this to remove mutation of the request body parameter, namely by creating a deep copy + assembling changes isolated to that copy.

That means switching to a top-level constant definition, with readonly properties.

Let's keep that for the next set of improvements.

src/controllers/did.ts
```diff
@@ -543,18 +565,21 @@ export class DIDController {
...
 		const identityServiceStrategySetup = new IdentityServiceStrategySetup(response.locals.customer.customerId);
+
 		try {
-			const did = request.params.did
+			const didDocument  = did
 				? await identityServiceStrategySetup.agent.resolveDid(request.params.did)
 				: await identityServiceStrategySetup.agent.listDids(response.locals.customer);
 
-			return response.status(StatusCodes.OK).json(did);
+			return response.status(StatusCodes.OK).json(didDocument satisfies ListDidsResponseBody | QueryDidResponseBody);
```
Eengineer1, line 578: There's a relationship between `ListDidsResponseBody`, `QueryDidResponseBody`, therefore I'd make those derived from each other, 1 is declared the other one references the other, with the latter `QueryDidResponseBody` being the declaration to reference from.

Eengineer1 (dismissed):
Looks good.

I've left several remarks as a set of non-blocking improvements to be considered for next release.

Approved.

PR #457, opened 2024-01-02
feat: Add did import api [DEV-3487]

DaevMithran:
This PR includes the addition of did/import api

At the moment it requires all the keys in the didDocument to be passed in the request, The ability to handle existing keys in wallet will be supported once the addition of that feature is complete in veramo, Created a feature request here https://github.com/decentralized-identity/veramo/issues/1312


ankurdotb:
Task linked: [DEV-3487 Add did/import api](https://app.clickup.com/t/6600954/DEV-3487)

src/services/identity/local.ts
```diff
@@ -122,11 +122,17 @@ export class LocalIdentityService extends DefaultIdentityService {
 		try {
 			return await this.getDid(ISSUER_DID);
 		} catch {
+			const key = {
+				kid: ISSUER_PUBLIC_KEY_HEX,
+				type: 'Ed25519' as TKeyType,
```
Eengineer1, line 127: Switch to the following for consistency:

```suggestion
				type: 'Ed25519' as const,
```

src/services/identity/agent.ts
```diff
@@ -244,21 +244,24 @@ export class Veramo {
 	async importDid(
 		agent: TAgent<IDIDManager>,
 		did: string,
-		privateKeyHex: string,
-		publicKeyHex: string
+		keys: Pick<IKey, 'privateKeyHex' | 'type'>[],
+		controllerKeyId: string
```
Eengineer1, line 248: I'd rather have the controller key identifier optional + assume the first verification method key id is the controller, if omitted.

Otherwise, if specified, it should be validated as a two step process:

- Validate if any of the included key ids, reject early if not.
- Bubble up generic import error.

src/services/identity/index.ts
```diff
@@ -74,7 +75,12 @@ export interface IIdentityService {
 	resolveDid(did: string): Promise<DIDResolutionResult>;
 	resolve(didUrl: string): Promise<Response>;
 	getDid(did: string, customer: CustomerEntity): Promise<any>;
-	importDid(did: string, privateKeyHex: string, publicKeyHex: string, customer: CustomerEntity): Promise<IIdentifier>;
+	importDid(
+		did: string,
+		keys: Pick<IKey, 'privateKeyHex' | 'type'>[],
+		controllerKeyId: string,
```
Eengineer1, line 81: Same logic applies here.

src/services/identity/postgres.ts
```diff
@@ -289,16 +290,15 @@ export class PostgresIdentityService extends DefaultIdentityService {
 
 	async importDid(
 		did: string,
-		privateKeyHex: string,
-		publicKeyHex: string,
+		keys: Pick<IKey, 'privateKeyHex' | 'type'>[],
+		controllerKeyId: string,
```
Eengineer1, line 294: Same logic applies here as well.

src/app.ts
```diff
@@ -167,6 +167,7 @@ class App {
 		// DIDs API
 		app.post('/did/create', DIDController.createDIDValidator, new DIDController().createDid);
 		app.post('/did/update', DIDController.updateDIDValidator, new DIDController().updateDid);
+		app.post('/did/import', new DIDController().importDid);
```
Eengineer1, line 170: This doesn't look right to me.

Where's the validation here?

I don't see any definition either.

Eengineer1 (changes requested):

src/app.ts
DaevMithran: Added validators for did/import and key/import. Thanks

src/services/identity/agent.ts
DaevMithran: Made the controllerKeyId optional

Validation of controllerKeyId and assigning the first one by default is handled in veramo @Eengineer1, should we do that here again.

Eengineer1 (approved):
Looks good.

Approved.

src/services/identity/agent.ts
Eengineer1: I'd add that comparison logic to the enclosed validator.

Leaving as separate future scope, as that's simple to implement if proven essential.

PR #481, opened 2024-01-24
feat: Allow DID controller switching [DEV-3587]

lampkin-diet:
- Add an ability to pass not the default keys to sign fr did update and deactivate and resource creation


tests/e2e/did/switch-controller-flow.spec.ts
```diff
@@ -0,0 +1,445 @@
...
+async function isDeactivated(did: string, request: APIRequestContext): Promise<boolean> {
+	const response = await request.get(`/did/search/${did}`);
+	expect(response).toBeOK();
+
+	return await response.json().then((result) => JSON.parse(result.didDocumentMetadata.deactivated));
+}
+
+function mustIncludeKey(didResponse: ResponseBody, keyRef: string): boolean {
+	return didResponse.controllerKeys.some((key) => key.kid === keyRef);
+}
```
Eengineer1, line 60: The naming convention is ambiguous here.

The expected `must<perform-action>` literal indicates the closure throws if the operation fails, which is not the case here.

I appreciate this is a test helper, therefore won't probably confuse end users, hence why marking as non-blocking to promote.

tests/e2e/did/switch-controller-flow.spec.ts
```diff
@@ -0,0 +1,445 @@
...
+
+	expect(response).toBeOK();
+	expect(response.status()).toBe(StatusCodes.OK);
+
+	// Make specific checks.
+	// 1. Check that the response contains all keys from the original DID and the new one.
+	const updatedDIDResponse = await response.json();
+	expect(updatedDIDResponse.keys.length).toBe(1);
+	expect(updatedDIDResponse.controllerKeys.length).toBe(1);
+	expect(mustIncludeKey(updatedDIDResponse, DID_A.controllerKeyId)).toBeTruthy();
```
Eengineer1, line 404: Marking as addressable as well.

tests/e2e/did/switch-controller-flow.spec.ts
```diff
@@ -0,0 +1,445 @@
...
+
+	// 2. ControllerKeyId field should be still the first verificationMEthod fron original controller
+	expect(updatedDIDResponse.controllerKeyId).toEqual(DID_A.controllerKeyId);
+
+	// 3. Check that each key has a controller field which points to the DID with it associated to.
+	expect(allControllerKeysHaveController(updatedDIDResponse)).toBeTruthy();
+
+	// 4. Check that the response contains all controller keyRefs.
+	expect(mustIncludeKeyControllerRef(updatedDIDResponse, DID_A.controllerKeyId)).toBeTruthy();
```
Eengineer1, line 413: Marking to be addressed.

src/services/identity/abstract.ts
```diff
@@ -85,7 +85,12 @@ export abstract class AbstractIdentityService implements IIdentityService {
 		throw new Error(`Not supported`);
 	}
 
-	createResource(network: string, payload: ResourcePayload, customer: CustomerEntity): Promise<any> {
+	createResource(
+		network: string,
+		payload: ResourcePayload,
+		customer: CustomerEntity,
+		publicKeyHexs?: string[]
```
Eengineer1, line 92: Adding as an all-inclusive reference to `publicKeyHexs` introduction to DID, DLR operations:

This assumes only keyrefs can be passed, therefore no external key importing is supported, by passing it as signature material information, otherwise known as `signInput`.

The functionality per se is exported only at the key / did import layer, which is sensible since we're running custodian mode for issuers, verifiers.

src/services/identity/postgres.ts
```diff
@@ -301,14 +328,27 @@ export class PostgresIdentityService extends DefaultIdentityService {
 		return identifier;
 	}
 
-	async createResource(network: string, payload: ResourcePayload, customer: CustomerEntity) {
+	async createResource(
```
Eengineer1, line 331: Shouldn't the local strategy be updated accordingly as well?

That's relevant for at least DLR creation, DID deactivation, ensuring consistent behavior on the intersection of inherently implemented or not methods.

Eengineer1 (changes requested):
Very good demonstration on TDD defining + addressing interconnected and / or interchangeable identifier update scenarios.

There are relatively minor remarks to resolve, which should be quick to modify, before that's ready to promote.

PS: Cool + exhaustive testing sequence!

tests/e2e/did/switch-controller-flow.spec.ts
```diff
@@ -0,0 +1,445 @@
...
+
+	// use data from the response
+	DID_C = (await response.json()) as ResponseBody;
+	expect(DID_C.controllerKeyId).toBeDefined();
+
+	DIDDocument_C = await searchDID(DID_C.did, request);
+	expect(DIDDocument_C).toBeDefined();
+
+	console.info(`DIDDocument-C: ${JSON.stringify(DIDDocument_C)}`);
+});
```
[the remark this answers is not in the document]
Eengineer1: Marking to be parallelised when we have exposed the async TX submission option in both `@cheqd/sdk`, `@cheqd/did-provider-cheqd`.

After v0.47 network upgrade, we will have optimistic submission available to work with as well.

PR #500, opened 2024-02-29
feat: Add base endpoints for Stripe integration [DEV-3671]

lampkin-diet:

PR #510, opened 2024-04-01
feat: Add API key endpoints [DEV-3793]

lampkin-diet:


ankurdotb:
Task linked: [DEV-3793 Implement endpoints for API keys](https://app.clickup.com/t/6600954/DEV-3793)

README.md
```diff
@@ -97,6 +97,9 @@ By default, `ENABLE_AUTHENTICATION` is set to off/`false`. To enable external Ve
     2. `LOGTO_WEBHOOK_SECRET`: Webhook secret to authenticate incoming webhook requests from LogTo.
 5. **Miscellaneous**
     1. `COOKIE_SECRET`: Secret for cookie encryption.
+    2. `API_KEY_PREFIX` (optional): Prefix for API keys. (Default "caas")
```
ankurdotb, line 100: Don't bother with making this settable. Do we need a prefix?

README.md
```diff
@@ -97,6 +97,9 @@ By default, `ENABLE_AUTHENTICATION` is set to off/`false`. To enable external Ve
...
+    3. `API_KEY_LENGTH` (optional): Length of API keys. (Default 32)
```
ankurdotb, line 101: Just use a default. No need to make this configurable

README.md
```diff
@@ -97,6 +97,9 @@ By default, `ENABLE_AUTHENTICATION` is set to off/`false`. To enable external Ve
...
+    4. `API_KEY_EXPIRATION` (optional): Expiration time for API keys in month. (Default 1 month)
```
ankurdotb, line 102: I would use days rather than months, since a "month" is more vague (is it 28? 29? 30? 31?)

ankurdotb (changes requested):
I haven't looked at the whole thing but left some comments based on the docs added

README.md
ankurdotb: What key length does Cloudflare use? I'd use similar (I believe they are either 40 or 64)

src/services/admin/api-key.ts
```diff
@@ -0,0 +1,169 @@
...
+
+	public static instance = new APIKeyService();
+
+	constructor() {
+		this.apiKeyRepository = Connection.instance.dbConnection.getRepository(APIKeyEntity);
+		this.secretBox = new SecretBox(process.env.EXTERNAL_DB_ENCRYPTION_KEY);
+	}
+
+	// ToDo: Maybe we also need to store not the API key but the hash of it?
+	// But in that case the API key will be shown once and then it will be lost.
```
Eengineer1, line 31: No reason to do so per se.

If handled properly, the user will save the secret on their end, along with our handling storing it safely encrypted, decrypting it in transit, if necessary.

If we're looking to compare the provided API key in headers, then we _should_ compute the HMAC SHA-256 / SHA-512 from the plain-text value **and** a user-specific unique identifier, such as the Logto short unique id or other.

The reconstructed value _must_ be calculated deterministically, in which case we _should_ store the same result in the DB on creation following the same process.

Eengineer1: The same can be achieved with bcrypt > 11 rounds.

Eengineer1:
Haven't changed the base to point to #500 but that would be easier to review if done so next time, declaring a merging sequence as well.

src/database/entities/api.key.entity.ts
```diff
@@ -56,11 +72,21 @@ export class APIKeyEntity {
...
+		apiKey: string,
+		name: string,
+		expiresAt: Date,
+		customer: CustomerEntity,
+		user: UserEntity,
+		revoked = false
+	) {
+		this.apiKeyHash = apiKeyHash;
 		this.apiKey = apiKey;
+		this.name = name;
```
DaevMithran, line 86: can we also add a `description` field, maybe an optional `url` as well.  It was required in the UI

DaevMithran (changes requested):

src/controllers/admin/api-key.ts
```diff
@@ -0,0 +1,420 @@
...
+		check('expiresAt')
+			.optional()
+			.isISO8601()
+			.toDate()
+			.withMessage('Invalid date format')
+			.custom((value) => {
+				if (value < new Date()) {
+					throw new Error('expiresAt must be in the future');
+				}
+				return true;
```
ankurdotb, line 40: AFAIK, `new Date()` will create a Date object based on the local timezone of the machine. Our server is in London, so it's actually just UTC for most of the year, except when it's summer time when it will be 1 hour out. The problem is more obvious if the server was in a completely different region/timezone, where this comparison could result in a failure.

You might need to something like the following (if you agree this could be a bug - maybe I'm misunderstanding this as something that could be a bug):

```typescript
.custom((value) => {
            const currentUtcDate = new Date(new Date().toISOString());
            if (value < currentUtcDate) {
                throw new Error('expiresAt must be in the future');
            }
            return true;
        })
```

Here, `toISOString` will create a normalised UTC timezone Date object

ankurdotb (changes requested):

src/middleware/auth/routes/admin/admin-auth.ts
```diff
@@ -48,6 +48,17 @@ export class AdminHandler extends BaseAuthHandler {
...
+		this.registerRoute('/admin/api-key/create', 'POST', 'admin:api-key:create:testnet', { skipNamespace: true });
+		this.registerRoute('/admin/api-key/create', 'POST', 'admin:api-key:create:mainnet', { skipNamespace: true });
+		this.registerRoute('/admin/api-key/update', 'POST', 'admin:api-key:update:testnet', { skipNamespace: true });
+		this.registerRoute('/admin/api-key/update', 'POST', 'admin:api-key:update:mainnet', { skipNamespace: true });
+		this.registerRoute('/admin/api-key/revoke', 'DELETE', 'admin:api-key:revoke:testnet', { skipNamespace: true });
+		this.registerRoute('/admin/api-key/revoke', 'DELETE', 'admin:api-key:revoke:mainnet', { skipNamespace: true });
+		this.registerRoute('/admin/api-key/get', 'GET', 'admin:api-key:get:testnet', { skipNamespace: true });
+		this.registerRoute('/admin/api-key/get', 'GET', 'admin:api-key:get:mainnet', { skipNamespace: true });
+		this.registerRoute('/admin/api-key/list', 'GET', 'admin:api-key:list:testnet', { skipNamespace: true });
+		this.registerRoute('/admin/api-key/list', 'GET', 'admin:api-key:list:mainnet', { skipNamespace: true });
```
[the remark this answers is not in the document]
ankurdotb: Addressed in #522

src/database/entities/api.key.entity.ts
ankurdotb: API key just needs a name, doesn't need a description. Organisation/customer might have a description, which is addressed in #522

src/controllers/admin/api-key.ts
lampkin-diet: Agreed!

ankurdotb (approved):

PR #515, opened 2024-04-05
fix: Restrict verida to mainnet and testnet

DaevMithran:


Eengineer1 (approved):
Looks good.

Approved.

PR #543, opened 2024-06-18
fix: Incorrect customer/Stripe details stored in DB [DEV-4005]

benyam7:


ankurdotb:
Task linked: [DEV-4005 Fix incorrect population of CustomerEntity in DB causing subscription check failure(Credential Service Backend)](https://app.clickup.com/t/6600954/DEV-4005)

src/controllers/api/account.ts
```diff
@@ -189,7 +189,9 @@ export class AccountController {
 			// Even if customer was created before for such user but the process was interruted somehow - we need to create it again
 			// Cause we don't know the state of the customer in this case
 			// 2.1.1. Create customer
-			customer = (await CustomerService.instance.create(logToUserEmail)) as CustomerEntity;
+			// I’m setting the "name" field to an empty string on the current CustomerEntity because it is non-nullable.
+			//  we will populate the customer's "name" field using the response from the Stripe account creation in account-submitter.ts.
+			customer = (await CustomerService.instance.create('', logToUserEmail)) as CustomerEntity;
```
ankurdotb, line 194: This `name` field should be the organisation/customer name that they set in settings. If you need a default to initialise (say, if the customer hasn't yet gone to that page and set it), could you instead set it to the `name` property of the Logto user? This might sometimes have a name, sometimes be blank. Better than just straight up defaulting to blank.

src/services/track/admin/account-submitter.ts
```diff
@@ -45,8 +45,8 @@ export class PortalAccountCreateSubmitter implements IObserver {
 				return;
 			}
 
-			// Update the CaaS customer with the new Stripe account
-			await CustomerService.instance.update(data.customerId, undefined, account.id);
+			// Update the CaaS customer with the new Stripe account. Note, we're populating the "name" field from stripe's response.
+			await CustomerService.instance.update(data.customerId, data.name, undefined, undefined, account.id);
```
ankurdotb, line 49: Bit opaque if you're just using `undefined` as positional parameters. Isn't it better to strongly type this, like so?

I assume `paymentProviderId` is what the value of `account.id` is.

```
await CustomerService.instance.update({
  customerId: data.customerId,
  name: data.name,
  paymentProviderId: account.id
});
```

ankurdotb (changes requested):

src/services/track/admin/account-submitter.ts
benyam7: Yeah, but `update` method in `CustomerService` doesn't take object. It is structured like this:

```
public async update(
		customerId: string,
		name?: string,
		email?: string,
		description?: string,
		paymentProviderId?: string
	)
```

src/controllers/api/account.ts
benyam7: Yeah, make sense. 👍🏾

benyam7: But, you it will get override by this bit in `account-submitter.ts` which uses name from stripe. But, it does make sense to keep the name from "stripe" as it is mandatory in checkout and "customer" table indeed referring to the stripe customer.
```
await CustomerService.instance.update(data.customerId, data.name, undefined, undefined, account.id);
```

src/services/track/admin/account-submitter.ts
benyam7: Should I refactor, the `update` method to take object instead?

ankurdotb: Yes, I think it might be more sensible to make it so we aren't relying on just ordered parameters being passed. What's the general style within this app: are such objects declared in types? Is there an existing type or combination of types that describes this object?

benyam7: For creating customers, we are relying on TypeORM declared entity called : `CustomerEntity`. In this case, the 'name' field is mandatory. Since updating requires fields to be partial, we cannot use that. So, I've created a concrete type in `types/customer.ts` to handle the update body.

ankurdotb: What you've done with a new `UpdateCustomerBody` definition is not _bad_ per se, but is not best practice, because you've just duplicated/replicated an existing definition somewhere else. Say the definition of the `CustomerEntity` itself changed in the future, someone would need to remember to update this `UpdateCustomerBody` and since they are not linked classes/interfaces in any form, unless they notice it they are very likely to miss it. (And if they miss it, new customer account creation may work but updates will fail).

What you're trying to achieve here is a class/type that is same definition as `CustomerEntity`, but with just `customerId` as mandatory. Here's a better way to achieve it:

```
import { CustomerEntity } from './path/to/customer-entity-definition';

export type UpdateCustomerEntity = Partial<CustomerEntity> & Pick<CustomerEntity, 'customerId'>;
```

You might need to play/check this works since it's a bit pseudo-code. You're using TypeScript, rely on its advantages! `Partial` allows you to inherit the parent with properties set to optional, and `Pick` allows you to set the selected property to mandatory. The union of those should result in a type that is similar to yours now. But the benefit is that if `CustomerEntity` is updated in the future, with different/new properties, `UpdateCustomerEntity` **does not** break because it's not a manual copy-paste of the properties.

You may also want to check whether there are any other bits that need to be updated, besides this type definition.

ankurdotb (approved):

PR #552, opened 2024-06-27
fix: weird issue

benyam7:
We're encountering a strange issue when checking if `request.body.user.name` exists. Even though, it should evaluate to `null` it is becoming an empty string.

This causes the app to fail because the `name` column must be unique, breaking the `Account bootstrap` webhook. Hence, breaking the whole account setup flow. 

Since we know the user won't have a `name` during the register flow, and even if they were to have a name we need to ensure `uniqueness`, hence using the `email` right away. Because, `email` uniqueness is already handled by logto.

### DIFF
--- a/src/controllers/api/account.ts
+++ b/src/controllers/api/account.ts
@@ -174,7 +174,10 @@ export class AccountController {
 		}
 		const logToUserId = request.body.user.id;
 		const logToUserEmail = request.body.user.primaryEmail;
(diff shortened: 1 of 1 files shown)


benyam7:
More on this [here](https://cheqd-identity.slack.com/archives/C01PTLM8RS6/p1719500071874509?thread_ts=1719483764.540119&cid=C01PTLM8RS6)

PR #561, opened 2024-07-25
feat: Synchronising stripe and studio API state - [DEV-4157]

jay-dee7:
This PR makes few adjustments and breaks existing assumptions about the user journeys for Stripe and our API.

- We now create a customer (Studio) when someone subscribes on Stripe but has never signed up for our platform. This helps with `Inbound Billing Journey`
- We still try to maintain consistency at runtime about 1-1 relationship between the `user` and `customer` tables. Though it's quite risky since the underlying database does not guarantee the relation terms. We could end up creating multiple customers in some edge case as long as it's not implemented at the database layer. The first step for this should be marking the `customer.email` as the primary or at least a unique column.
- Allow setting the `paymentProviderId` while creating new Customer entity records, it was optional before anyway and we only allow to create it optionally (to avoid having to perform a create + update operation)

- [x] New endpoint - `GET /admin/checkout/session/:id` API to retrieve Stripe checkout session by ID

>[!WARNING]
> ~~The get Stripe checkout session endpoint is completely open at the source (API) this of course does require some thinking but the times when this endpoint becomes useful is mostly when user is not authenticated. One thing we could do is make this `authenticated` and force a user login before performing any checks on client side.~~


src/services/track/admin/subscription-submitter.ts
```diff
@@ -6,6 +7,7 @@ import { CustomerService } from '../../api/customer.js';
 import type { ISubmitOperation, ISubmitSubscriptionData } from '../submitter.js';
 import { EventTracker } from '../tracker.js';
 import type { IObserver } from '../types.js';
+import { FindOptionsWhere } from 'typeorm';
```
DaevMithran, line 10: This is showing an error. 

```suggestion
import type { FindOptionsWhere } from 'typeorm';
```

DaevMithran:
`SubscriptionSubmitter -> update` function is not used anywhere. Can we remove it?

.node-version
```diff
@@ -0,0 +1 @@
+20
```
DaevMithran, line 1: is this change needed for stripe?

jay-dee7: No not really. This project didn't build or run with any other node version initially so I just added this to make it easier to work with node versions. But now that I see, it builds fine with even the latest version so this might not be needed.

jay-dee7:
> @jay-dee7 `SubscriptionSubmitter -> update` function is not used anywhere. Can we remove it?

@DaevMithran we can't remove it. To implement `IObserver` interface we only need to satisfy the update method on it so it's required in implementation. Of course we might still not be using it. I'd just keep it out of scope for now.

DaevMithran (approved):

ankurdotb:
Task linked: [DEV-4157 Create flow for external subscription through static pricing table](https://app.clickup.com/t/6600954/DEV-4157)

PR #565, opened 2024-08-01
feat: Switch logto role assignment to subscription webhook

jay-dee7:


src/controllers/api/account.ts
```diff
@@ -301,41 +394,9 @@ export class AccountController {
...
 
-		const logToHelper = new LogToHelper();
-		const _r = await logToHelper.setup();
-		if (_r.status !== StatusCodes.OK) {
-			return response.status(StatusCodes.BAD_GATEWAY).json({
-				error: _r.error,
-			} satisfies UnsuccessfulResponseBody);
-		}
 		// 5. Assign default role on LogTo
 		// 5.1 Get user's roles
```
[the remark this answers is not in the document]
Eengineer1: Same applies for each removed / replaced block of actions.

src/middleware/auth/logto-helper.ts
```diff
@@ -287,58 +258,104 @@ export class LogToHelper extends OAuthProvider implements IOAuthProvider {
...
+		const [userInfo, userRoles] = await Promise.all([this.getUserInfo(userId), this.getRolesForUser(userId)]);
+		console.log('assignCustomerPlanRoles: userInfo: ', userInfo);
+		console.log('assignCustomerPlanRoles: userRoles: ', userRoles);
 
 		if (userInfo.status !== StatusCodes.OK) {
-			return this.returnError(
-				StatusCodes.BAD_GATEWAY,
-				`Could not fetch the info about role with roleId ${roleId}`
-			);
+			return this.returnError(StatusCodes.BAD_GATEWAY, 'Could not fetch the info about the user');
```
Eengineer1, line 267: The status code doesn't look right to me.

Again, 503 would be better placed vs 504.

src/middleware/auth/logto-helper.ts
```diff
@@ -287,58 +258,104 @@ export class LogToHelper extends OAuthProvider implements IOAuthProvider {
...
 		}
 		// Means that user exists
-		if (userInfo.data.isSuspended === 'true') {
+		if (userInfo.data.isSuspended) {
```
Eengineer1, line 270: Let's be mindful of the underlying type and / or possible coercion in effect here.

Otherwise, this might be an always true condition.

src/middleware/auth/logto-helper.ts
```diff
@@ -287,58 +258,104 @@ export class LogToHelper extends OAuthProvider implements IOAuthProvider {
...
-		// Means it's not suspended
-		const role = await this.getRoleInfo(roleId);
-		if (role.status !== StatusCodes.OK) {
-			return this.returnError(
-				StatusCodes.BAD_GATEWAY,
-				`Could not fetch the info about user with userId ${userId} because of error from authority server: ${role.error}`
-			);
+
+		if (userRoles.status !== StatusCodes.OK) {
+			return this.returnError(StatusCodes.BAD_GATEWAY, 'Could not fetch the info about user roles');
```
Eengineer1, line 275: Same logic applies here regarding the status code returned.

src/middleware/auth/logto-helper.ts
```diff
@@ -287,58 +258,104 @@ export class LogToHelper extends OAuthProvider implements IOAuthProvider {
...
+		// All of our users get a "Portal" role which let's them work with our Studio Portal (UI)
+		// Test plan lets you operate with our Testnet so we assign the "Testnet" role
+		// Build plan lets you work with our Testnet and Mainnet, so we assign "Testnet" and "Mainnet" roles
+		// "build" is the superset of all the roles (testnet + mainnet)
+		const buildPlanRoleIds = [env.LOGTO_MAINNET_ROLE_ID.trim(), env.LOGTO_TESTNET_ROLE_ID.trim()];
+		const testPlanRoleId = env.LOGTO_TESTNET_ROLE_ID.trim();
+
+		const planRoleIds: string[] = [];
+		console.log('planType: ', planType);
+		if (planType === 'build') {
```
Eengineer1, line 291: Let's use a constant literal object definition here instead.

src/middleware/auth/logto-helper.ts
```diff
@@ -287,58 +258,104 @@ export class LogToHelper extends OAuthProvider implements IOAuthProvider {
...
+			const buildRoleIsAssigned = buildPlanRoleIds.every((roleId) => assignedRoleIds.includes(roleId));
+			if (buildRoleIsAssigned) {
+				return {
+					status: 201,
+					error: '',
+					data: {
+						message: `${planType} plan role was successfull assigned to the user`,
```
Eengineer1, line 298:
```suggestion
						message: `${planType} plan role was successfully assigned to the user`,
```

src/middleware/auth/logto-helper.ts
```diff
@@ -287,58 +258,104 @@ export class LogToHelper extends OAuthProvider implements IOAuthProvider {
...
+				await this.removeLogtoRoleFromUser(userId, mainnetRoleId);
+			}
+
+			const testnetRoleId = assignedRoleIds.find((roleId) => roleId === testPlanRoleId);
+			if (testnetRoleId) {
+				return {
+					status: 201,
+					error: '',
+					data: {
+						message: `${planType} plan role was successfull assigned to the user`,
```
Eengineer1, line 320:
```suggestion
						message: `${planType} plan role was successfully assigned to the user`,
```

Eengineer1 (changes requested):

src/services/track/admin/subscription-submitter.ts
```diff
@@ -36,14 +40,119 @@ export class SubscriptionSubmitter implements IObserver {
...
+
+		if (user) {
+			switch (operation.operation) {
+				case OperationNameEnum.SUBSCRIPTION_CREATE:
+				case OperationNameEnum.SUBSCRIPTION_UPDATE:
+					this.handleCutomerRoleAssignment(operation, logToHelper, user.logToId, productName);
+					return;
+				case OperationNameEnum.SUBSCRIPTION_CANCEL:
+					this.handleCustomerRoleRemoval(operation, logToHelper, user.logToId);
+					return;
```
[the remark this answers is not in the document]
jay-dee7: @Eengineer1 subscription cancellation is immediate here as well (just like creds). We issue a prorate refund immediately.

src/services/track/admin/subscription-submitter.ts
```diff
@@ -126,6 +236,8 @@ export class SubscriptionSubmitter implements IObserver {
 
 	async submitSubscriptionUpdate(operation: ISubmitOperation): Promise<void> {
 		const data = operation.data as ISubmitSubscriptionData;
+		const stripe = new Stripe(process.env.STRIPE_SECRET_KEY);
```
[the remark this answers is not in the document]
jay-dee7: I was just trying to use the existing methodologies. But I think it makes sense to move this into the Class property level and initiate once inside the constructor?

src/controllers/api/account.ts
```diff
@@ -175,18 +244,28 @@ export class AccountController {
...
 			UserService.instance.get(logToUserId),
 			CustomerService.instance.find({ email: logToUserEmail }),
+			logToHelper.setup(),
 		]);
 
+		if (_r.status !== StatusCodes.OK) {
+			return response.status(StatusCodes.BAD_GATEWAY).json({
+				error: _r.error,
+			} satisfies UnsuccessfulResponseBody);
+		}
```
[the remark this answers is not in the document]
jay-dee7: I think that sense, I've switched to `503 Service Unavailable`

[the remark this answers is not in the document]
jay-dee7: Also changed the shorthand naming to be a bit more readable and switched overlapping names to indicate where they come from, i.e customer -> `customerEntity` and stripe customer becomes `stripeCustomer`

src/middleware/auth/logto-helper.ts
jay-dee7: Wouldn't it be actually better to just proxy the actual response status?

src/middleware/auth/logto-helper.ts
jay-dee7: I've checked it, this is a boolean. The existing condition would never has worked out and the reason why it went unnoticed is because there's no typings available on most of the responses in the project. It's an `any` here.

src/middleware/auth/logto-helper.ts
jay-dee7: I've converted this to a `String Enum` type and then compare using enum variants

src/middleware/auth/logto-helper.ts
jay-dee7: I've switched to underlying status code and changed the response code in the request sender wrapper to use `503 Service Unavailable` when we actually can't reach it

ankurdotb:
Task linked: [DEV-4164 fix: align the LogTo roles with the cheqd Studio plans](https://app.clickup.com/t/6600954/DEV-4164)

.node-version
ankurdotb, line 1: Necessary?

src/controllers/api/account.ts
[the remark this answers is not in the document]
ankurdotb: Is this resolved?

src/middleware/auth/logto-helper.ts
ankurdotb, line 1: This is mostly fine but I think funny stuff will happen if the user is on a custom plan. If the user isn't on one of the known plans for test/build (say custom plan), do nothing to roles as this will likely then get done manually.

src/services/api/role.ts
```diff
@@ -57,6 +58,6 @@ export class RoleService {
 	}
 
 	public getDefaultRole() {
-		return this.roleRepository.findOneBy({ name: 'default' });
```
ankurdotb, line 60: I suspect this is the reason why your tests are failing. LogTo role `default` is a special type of role AFAIK and it was used for unauthenticated user tests, i.e., any APIs users could use without being logged in at all. (We did have such APIs initially, e.g., DID search, DID get which was basically DID resolution.) I don't think we have this any more.

ankurdotb: Huh, I checked the value of the const and not what I was expecting it to be. 🤔

.node-version
jay-dee7: This is actually necessary since this project only works with node 20 because we use import statements with `assert` keyword. This has been since replaced with `with` keyword but even some project deps use this so we're stuck with this

src/services/api/role.ts
jay-dee7: This one is only fetching it from the database. We do store logto role ids in DB too and this one is a helper method to find the default role

.node-version
ankurdotb: Isn't this handled by the `engines: node` configuration in package.json? Or does setting a .node-version make it behave like `engines-strict`? (I don't think it does the latter as far as I'm aware.) Not sure if a global config for Node on your machine is causing issues.

jay-dee7: well actually engines property would work as well but we need to pin it down to a version (something like `20.11.0`) instead of >=.
This was only really for my convenience since my local tooling auto-detects this and switches node version

src/middleware/auth/logto-helper.ts
jay-dee7: We don't actually remove any role outside of testnet or mainnet roles. So any role attached manually that isn't  mainnet or testnet will not be removed in this instance

.node-version
ankurdotb: Then the correct change here would be to changes the `engines` property in package.json rather than introducing the node-version file. I've made this change.

ankurdotb (approved):

PR #567, opened 2024-08-06
feat: Provision mainnet account for customers on account bootstrap [DEV-4203]

jay-dee7:


ankurdotb:
Task linked: [DEV-4203 cheqd Studio: fix mainnet account creation for both staging and prod](https://app.clickup.com/t/6600954/DEV-4203)

src/controllers/api/account.ts
```diff
@@ -288,54 +287,48 @@ export class AccountController {
...
 
+		const mainnetAccount = mainnetAccountResponse.data;
+		const testnetAccount = testnetAccountResponse.data;
+
 		// 7. Assign default role on LogTo
 		const customDataFromLogTo = await logToHelper.getCustomData(userEntity.logToId);
 
 		// 8. Create custom_data and update the userInfo (send it to the LogTo)
-		if (Object.keys(customDataFromLogTo.data).length === 0 && paymentAccount.address) {
+		if (Object.keys(customDataFromLogTo.data).length === 0 && testnetAccount.address && mainnetAccount.address) {
```
DaevMithran, line 320: if the mainnet account address is undefined we can set it as null in the custom data? useful for staging environment

DaevMithran (changes requested):

DaevMithran (changes requested):
`provisionCustomerAccount` should also be used in `/account/create` api [here](https://github.com/cheqd/studio/blob/feat/mainnet-account-provision-on-bootstrap/src/controllers/api/account.ts#L502).

DaevMithran:
Lets return both the payment addresses in `GET /account` api [here](https://github.com/cheqd/studio/blob/feat/mainnet-account-provision-on-bootstrap/src/controllers/api/account.ts#L83)

src/controllers/api/account.ts
```diff
@@ -381,6 +374,69 @@ export class AccountController {
 		return response.status(StatusCodes.OK).json({});
 	}
 
+	static async provisionCustomerAccount(
+		network: CheqdNetwork,
+		accounts: PaymentAccountEntity[],
+		customerEntity: CustomerEntity
+	): Promise<SafeAPIResponse<PaymentAccountEntity>> {
+		const existingAccount = accounts.find((acc) => acc.namespace === network);
```
DaevMithran, line 382: Do we need to check for mainnet permissions before creating the mainnet account?

jay-dee7: No @DaevMithran we create testnet and mainnet accounts for all users

src/controllers/api/account.ts
jay-dee7: No this will work on staging as well, it's all plan dependant now. You get two accounts (testnet and mainnet) and you operate using them with a plan (test or build)

src/controllers/admin/organisation.ts
```diff
@@ -83,7 +85,12 @@ export class OrganisationController {
 				name: customer.name,
 				email: customer.email,
 				description: customer.description,
-				cosmosAddress: paymentAccount[0].address as string,
+				cosmosAddress: {
+					[CheqdNetwork.Testnet]: paymentAccount.find((acc) => acc.namespace === CheqdNetwork.Testnet)
+						?.address,
```
DaevMithran, line 90: lets return `null` if not found

src/controllers/api/account.ts
DaevMithran: Okay, then I don't think these conditions are necessary. both mainnet and test account addresses are never undefined

DaevMithran (dismissed):

ankurdotb (dismissed):

ankurdotb (approved):

PR #568, opened 2024-08-06
perf: Parallelize non ledger operation tests

DaevMithran:
This PR separates tests into parallel and sequential folders, sequential tests are only for ledger operations.


playwright.sequential.config.ts
```diff
@@ -0,0 +1,102 @@
+import { defineConfig, devices } from '@playwright/test';
+import { STORAGE_STATE_AUTHENTICATED } from './tests/e2e/constants';
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+export default defineConfig({
```
rlve, line 7: To make it cleaner, instead of duplicating the config, you could build it dynamically for parallel/sequential run (i.e. based on some env var)

rlve (dismissed):

playwright.sequential.config.ts
DaevMithran: Hey @rlve I wanted to confirm, is it possible to run the parallel and sequential folders using a single server..

- I tried setting reuseExistingServer: true and running both of them parallely but it did not work.
- Is it possible to do this using a single file?, I was not able to set custom worker limit for a specific project
- I also tried running them in different ports but I am receiving weird [errors](https://github.com/cheqd/studio/actions/runs/10268957619/job/28413261551)

rlve: Yeah, I think there is an option, although might be a bit messy.

You would need to put all the sequential scenarios in one file and add this at the beginning:
`test.describe.configure({ mode: 'serial' });`

And then use only parallel config. The above would handle sequential scenarios.

DaevMithran: @rlve I'm using one playwright file now which uses an env variable to decide parallel vs sequential.

rlve (approved):

PR #572, opened 2024-08-22
feat: Allow additional data in credential request [DEV-4257]

DaevMithran:


ankurdotb:
Task linked: [DEV-4257 Add "termsOfUse" as a supported top level field for VCs](https://app.clickup.com/t/6600954/DEV-4257)

src/types/credential.ts
```diff
@@ -25,6 +25,17 @@ export interface CredentialRequest {
 	credentialSummary?: string;
 }
 
+export type CredentialRequest = BaseCredentialRequest & {
```
DaevMithran, line 28: is this approach fine? or can we allow any arbitrary data.

[x: string]: any

PR #574, opened 2024-08-23
fix: Fix customer creation workflow [DEV-4287]

DaevMithran:
- [x] Make email mandatory and unique in Cusomter entity
- [x] Move subscription-submitter logic within webhook handler
- [x] Dont add email in where clause if customer already has a paymentProviderId


ankurdotb:
Task linked: [DEV-4287 fix: Onboarding flow is not properly provisioning accounts](https://app.clickup.com/t/6600954/DEV-4287)

src/database/migrations/MigrateData.ts
```diff
@@ -99,7 +99,7 @@ export class MigrateData1695740345977 implements MigrationInterface {
 		for (const oldCustomer of await queryRunner.query(`SELECT * FROM customers`)) {
 			// 1. Create customer row:
 			console.info(`Creating CustomerEntity with id ${oldCustomer.customerId}`);
-			const customerEntity = new CustomerEntity(uuidv4(), oldCustomer.address);
```
ankurdotb, line 102: Is this "address" the cheqd address? If yes then I agree about switching to emails since it's more unique, whereas user might have more than one address (one for mainnet, one for testnet)

src/controllers/api/account.ts
```diff
@@ -480,14 +481,14 @@ export class AccountController {
...
-		const { username } = request.body;
+		const { name, primaryEmail } = request.body;
 
 		try {
 			// 2. Check if the customer exists
 			const customerEntity = response.locals.customer
 				? (response.locals.customer as CustomerEntity)
 				: // 2.1 Create customer
-					((await CustomerService.instance.create(username)) as CustomerEntity);
+					((await CustomerService.instance.create(name || primaryEmail, primaryEmail)) as CustomerEntity);
```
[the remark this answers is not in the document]
DaevMithran: Yes @ankurdotb. 

we have `UserEntity` which establishes the relationship between customer and logtoUserId.
This PR is replacing the uniqueness check with email instead of name.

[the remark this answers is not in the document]
DaevMithran: But the stripe email being same as the logto email / customer email will lead to a failure.

src/services/track/admin/subscription-submitter.ts
```diff
@@ -152,37 +152,40 @@ export class SubscriptionSubmitter implements IObserver {
 				this.stripe.customers.retrieve(data.paymentProviderId),
 			]);
 			if (!customer) {
-				const whereClause: FindOptionsWhere<CustomerEntity>[] = [{ paymentProviderId: data.paymentProviderId }];
-				// we add an additional "OR" check in case that a customer was created locally with email and no paymentProviderId
-				if (!stripeCustomer.deleted && stripeCustomer.email) {
-					whereClause.push({ email: stripeCustomer.email });
-				}
-
 				const customers = await CustomerService.instance.customerRepository.find({
```
DaevMithran, line 155: @jay-dee7 Can we just use findByPaymentProviderId here. Do we need to check for duplicate paymentProvidersId at this step? I think we can do that in a customer created event right?

src/controllers/api/account.ts
```diff
@@ -480,14 +481,14 @@ export class AccountController {
 		// 4. Check the token balance for Testnet account
 
 		// 1. Get logTo UserId from request body
-		const { username } = request.body;
+		const { name, primaryEmail } = request.body;
```
jay-dee7, line 484: @DaevMithran the `createValidator` is out of sync with current implementation.

src/controllers/admin/webhook.ts
```diff
@@ -5,10 +5,24 @@ import { StatusCodes } from 'http-status-codes';
...
 export class WebhookController {
+	private readonly stripe: Stripe;
+
+	constructor() {
+		this.stripe = new Stripe(process.env.STRIPE_SECRET_KEY);
+	}
+
+	public static instance = new WebhookController();
+
 	public async handleWebhook(request: Request, response: Response) {
```
jay-dee7, line 26: We don't need to initialise the Stripe client here since it's already initialised in two other places.
1. HTTP Handlers - `response.locals.stripe`
2. Constructor of this -> `this.stripe`

src/controllers/admin/webhook.ts
```diff
@@ -43,69 +57,356 @@ export class WebhookController {
...
+				case 'customer.subscription.trial_will_end':
+					subscription = event.data.object;
+					status = subscription.status;
+					await eventTracker.notify({
+						message: EventTracker.compileBasicNotification(
+							`Subscription status is ${status} for subscription with id: ${subscription.id}`,
+							'Stripe Webhook: customer.subscription.trial_will_end'
+						),
+						severity: 'info',
+					} satisfies INotifyMessage);
```
jay-dee7, line 73: @DaevMithran should we not perform some action on this one? if not, then might as well remove this?

src/controllers/admin/webhook.ts
```diff
@@ -43,69 +57,356 @@ export class WebhookController {
...
+					subscription = event.data.object;
+					status = subscription.status;
+					await eventTracker.notify({
+						message: EventTracker.compileBasicNotification(
+							`Subscription status is ${status} for subscription with id: ${subscription.id}`,
+							'Stripe Webhook: customer.subscription.deleted'
+						),
+						severity: 'info',
+					} satisfies INotifyMessage);
+					await this.handleSubscriptionCancel(subscription);
```
jay-dee7, line 85: A general style recommendation -
Since you're already handling the entire logic in these `handleSubscription<XXXX>` helper functions, it might be easier to just move the entire section inside the helper function and return from here?
- No need to declare a variable outside of switch block
- Notification can be already printed from inside the return
- You can just `return` instead of break and return

jay-dee7 (changes requested):
@DaevMithran the overall work looks good. Let's resolve all the comments and then we're good

src/controllers/admin/webhook.ts
jay-dee7, line 1: @DaevMithran also, since we've changed the way we handle webhooks from stripe now, it might (or might not) affect the client side behaviour, please test this end-to-end thoroughly as well.

src/controllers/api/account.ts
DaevMithran: Fixed it

src/controllers/admin/webhook.ts
DaevMithran: Removed

src/controllers/admin/webhook.ts
DaevMithran: @jay-dee7 added your first 2 suggestions. Leaving the third one as it is. As there are other parts of the code without response using the handlers

src/controllers/admin/webhook.ts
DaevMithran: @jay-dee7 Tested by triggering webhooks

PR #590, opened 2024-09-10
fix: Fix error handling in webhook

DaevMithran:


src/controllers/admin/webhook.ts
```diff
@@ -280,7 +280,7 @@ export class WebhookController {
 
 		const customer = await CustomerService.instance.findbyPaymentProviderId(data.paymentProviderId);
 		if (customer) {
-			this.syncLogtoRoles(operation, customer.customerId, '');
+			await this.syncLogtoRoles(operation, customer.customerId, '');
```
DaevMithran, line 283: should the product name be an empty string? @jay-dee7

jay-dee7: This is only valid for Cancel subscription.

jay-dee7 (approved):
LGTM @DaevMithran 🥇 🥇 🥇

PR #643, opened 2025-03-04
fix: Changes for resolver updates [DEV-4772]

sownak:


DaevMithran (dismissed):
Would be good to add a supporting test. But the changes are approved

PR #658, opened 2025-04-02
fix: Fix expiration date validation

DaevMithran:


Eengineer1 (approved):
Looks good.

Approved.

PR #662, opened 2025-04-08
ci: Run e2e tests in localnet

DaevMithran:
Need to update tesnet rpc url, external db url pointed to localnet


DaevMithran:
This is not needed

PR #673, opened 2025-05-05
perf: Improve customer account query [DEV-4965]

DaevMithran:


ankurdotb:
Task linked: [DEV-4965 cheqd Studio: Add index for customers to fix long transaction times](https://app.clickup.com/t/6600954/DEV-4965)

src/types/environment.d.ts
```diff
@@ -8,6 +8,7 @@ declare global {
 			PORT: string;
 			NPM_CONFIG_LOGLEVEL: string;
 			LOG_LEVEL: string | 'info';
+			LOCAL_STORE_TTL: string | undefined;
```
Eengineer1, line 11: As mentioned under PR #674.

src/database/entities/payment.account.entity.ts
```diff
@@ -8,6 +8,7 @@ import type { KeyEntity } from './key.entity.js';
 dotenv.config();
 
 @Entity('paymentAccount')
+@Index(['customer', 'namespace'], { unique: true })
```
Eengineer1, line 11: Looks like generated migrations are not saved under `src/database/migrations`, if one runs `npm run generate`.

Could you please resolve, generate and run migrations on staging?

I do remember migrations being run within DO. If that's the case, skip running them manually.

Eengineer1 (changes requested):

src/types/environment.d.ts
DaevMithran: Added this env in the Docker file

src/database/entities/payment.account.entity.ts
DaevMithran: Added the migration script

Eengineer1 (approved):
Approved 👍.

PR #674, opened 2025-05-06
perf: Cache customer accounts [DEV-4973]

DaevMithran:


ankurdotb:
Task linked: [DEV-4973 cheqd Studio: Implement caching of keyIDs](https://app.clickup.com/t/6600954/DEV-4973)

src/types/environment.d.ts
```diff
@@ -8,6 +8,7 @@ declare global {
 			PORT: string;
 			NPM_CONFIG_LOGLEVEL: string;
 			LOG_LEVEL: string | 'info';
+			LOCAL_STORE_TTL: string | undefined;
```
Eengineer1, line 11: Don't we need this set in the server / container config as well?

You're overriding during runtime, but ideally we want this explicitly set.

Eengineer1 (approved):
Approved, with non-blocking remark to address.

PR #698, opened 2025-06-20
fix: Allow tracking operations without resources [DEV-5109]

filipdjokic:
- `PaymentEntity`: Changed resource property type from `ResourceEntity` to `ResourceEntity | null`
- `PaymentService`: Updated `create()` and `update() `method signatures to accept `ResourceEntity | null`
- `TrackOperationWithPayment`: Fixed `getResourceEntity()` to use proper type guards instead of hard-coded property checks
- `OperationSubscriber`: Removed error when no resource is found (since it's normal for some operations)

- Updated `release` test with `getServiceEndpointValue ` to comply with [this DID resolver fix](https://github.com/cheqd/did-resolver/commit/bc108be5f7c155e34d043bce227c44e9d93d6de2)

### DIFF
--- a/src/database/entities/payment.entity.ts
+++ b/src/database/entities/payment.entity.ts
@@ -43,9 +43,9 @@ export class PaymentEntity {
 	@JoinColumn({ name: 'customerId' })
 	customer!: CustomerEntity;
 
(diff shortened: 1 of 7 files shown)


Eengineer1 (approved):
Changes are sensible. Type modifications look safely handled.

Approved 👍.

PR #710, opened 2025-08-05
feat: Include user scopes in idToken strategy

DaevMithran:
### Issue
The portal token strategy only considered scopes from the M2M token. Even if a user had the necessary permissions (scopes) assigned in Logto, these were not included in the effective scopes during API authorization.

### Fix
This PR updates the authentication strategy to:
   - Extract user-specific scopes from the ID Token payload.
   - Merge scopes from both the M2M token and the ID Token


Eengineer1 (approved):
Looks good.

Approved.

PR #725, opened 2025-09-04
feat: external provider support [DEV-5219]

sownak:
External Provider configurations
- Includes two new tables CredentialProvider for supported providers, and providerConfiguration for per customer provider config
- activate (to activate with new keys without redirecting to their website
- delete is for removing the keys and reseting provider config
- test (to test the connection)
- activated returns all activated providers


src/database/entities/provider-configuration.entity.ts
```diff
@@ -0,0 +1,123 @@
...
+	@PrimaryGeneratedColumn('uuid')
+	configId!: string;
+
+	@Column({
+		type: 'varchar',
+		nullable: false,
+	})
+	providerId!: string;
+
+	@Column({
```
DaevMithran, line 20: We can achieve the functionality by storing the encrypted api key in the existing api key table with a new column added for providerId.

src/database/entities/credential-provider.entity.ts
```diff
@@ -0,0 +1,120 @@
...
+	@Column({
+		type: 'varchar',
+		nullable: false,
+	})
+	providerType!: string; // 'credential-all', 'credential-issuance', 'credential-verification', 'etc'
+
+	@Column({
+		type: 'json',
+		nullable: false,
+		default: '[]',
```
DaevMithran, line 36: A lot of these fields can be part of the implementation of the provider.

src/services/api/dock-api.service.ts
```diff
@@ -0,0 +1,105 @@
+import * as dotenv from 'dotenv';
+import { ProviderService } from './provider.service.js';
+
+dotenv.config();
+
```
DaevMithran, line 5: All the providers should have to implement a common interface similar to our credo-ts, acapy work.

```
export interface IProviderService {
public readonly supportedProvider = 'dock';
  public static instance = new NewProviderApiService('NEW_PROVIDER_MASTER_API_KEY');

  protected getDefaultApiUrl(): string {
    return 'https://api.newprovider.com';
  }

```

src/database/entities/provider-configuration.entity.ts
DaevMithran: The api key can be encrypted by the function present already in the api key service file.

src/database/entities/provider-configuration.entity.ts
```diff
@@ -0,0 +1,123 @@
...
+		type: 'text',
+		nullable: false,
+	})
+	encryptedApiKey!: string;
+
+	@Column({
+		type: 'text',
+		nullable: false,
+	})
+	apiEndpoint!: string;
```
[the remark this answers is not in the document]
sownak: Agreed, but WebhookURL is customer/receiver specific, so I will keep it. Its not provider webhook, will revisit when webhook implementation

src/database/entities/provider-configuration.entity.ts
```diff
@@ -0,0 +1,123 @@
...
+	@Column({
+		type: 'timestamptz',
+		nullable: true,
+	})
+	validatedAt?: Date;
+
+	@Column({
+		type: 'timestamptz',
+		nullable: true,
+	})
```
DaevMithran, line 53: Is healthcheck related to a customer?

src/database/entities/provider-configuration.entity.ts
```diff
@@ -0,0 +1,123 @@
...
+	webhookUrl?: string;
+
+	@Column({
+		type: 'boolean',
+		default: false,
+	})
+	validated!: boolean;
+
+	@Column({
```
DaevMithran, line 44: Not required

src/database/entities/credential-provider.entity.ts
```diff
@@ -0,0 +1,120 @@
...
+		type: 'text',
+		nullable: true,
+	})
+	documentationUrl?: string;
+
+	@Column({
+		type: 'text',
+		nullable: true,
+	})
+	apiUrl?: string;
```
DaevMithran, line 70: Each provider has apiUrls for different enivronments. I suggest to skip this field for now

src/database/entities/credential-provider.entity.ts
```diff
@@ -0,0 +1,120 @@
...
+	})
+	capabilities!: string[];
+
+	@Column({
+		type: 'text',
+		nullable: true,
+	})
+	logoUrl?: string;
+
+	@Column({
```
DaevMithran, line 60: We can skip the fields and retain only 

- providerId
- name
- createdAt
- updatedAt 
- deprecated

src/database/entities/provider-configuration.entity.ts
sownak: subaccount, tenantid etc is abstracted inside the defaultSettings column which is json value

src/database/entities/credential-provider.entity.ts
sownak: The provider apiURL is default URL, it can be updated for each customer, so test/mainnet urls can be customized

PR #729, opened 2025-09-16
feat: Init dock integration [DEV-5224]

DaevMithran:


ankurdotb:
Task linked: [DEV-5224 Support credential API connectors for Dock](https://app.clickup.com/t/6600954/DEV-5224)

src/services/api/providers/provider.service.ts
```diff
@@ -0,0 +1,198 @@
+import { Repository } from 'typeorm';
```
sownak, line 1: Duplicate file, remove the other file.

src/services/api/providers/provider.factory.ts
```diff
@@ -0,0 +1,50 @@
+import { IProviderService } from '../../../types/provider.types';
```
sownak, line 1: Duplicate file, remove the other file

PR #731, opened 2025-09-22
fix: UI bugs and default assertionMethod [DEV-5287]

sownak:
- fix: Error in /credential/list when any one provider is not active

- New cheqd sdk version to support default assertionMethod


ankurdotb:
Task linked: [DEV-5287 Add default assertion method in `did/create` API to unlock JSON-LD issuance & verification](https://app.clickup.com/t/6600954/DEV-5287)

PR #742, opened 2025-10-06
fix: dock credential issuance check

sownak:


src/controllers/api/credential.ts
```diff
@@ -739,10 +737,8 @@ export class CredentialController {
 							new IdentityServiceStrategySetup(response.locals.customer.customerId).agent.listCredentials(
 								{
 									filter,
-									page,
-									limit: limit
-										? limit / activeProviderIds.length
-										: Math.floor(10 / activeProviderIds.length),
+									page: 1,
+									limit: 1000, // Use a large limit to get all data
```
DaevMithran, line 741: @sownak I think we can use the previous condition. But a fix if the limit is less than 0 then set it to 1? cuz we can send a hardcoded limit of 1000 in the UI and not hardcode it here

sownak: ok, will replace with the limit sent by UI. The previous condition was wrong as it was evenly distributing the limit, ie if there are 3 providers, limit for each will be 10/3 = 3.

sownak: I have added a multiplier for the limit

PR #810, opened 2026-04-09
feat: Add Faucet API access headers

filipdjokic:
This PR adds CF Zero Trust headers for communication with the Testnet Faucet API.

Not blocking and not dependent on any faucet modifications.


Eengineer1 (approved):
Looks good.

Approved 👍.

PR #811, opened 2026-05-17
feat: Align Basic/Build plans and add testnet faucet API

Eengineer1:
## Summary
- add Basic/Build plan capability mapping and keep the legacy test plan only as a fallback
- add authenticated `POST /account/faucet` access for Studio users with cap, balance, and request-more response details
- update subscription activation/migration docs and env examples for Basic/Build plan IDs
- add focused unit coverage for plan capabilities

## Verification
- `npm run build:app`
- `npm run test:unit -- tests/unit/admin/plan-capabilities.test.ts --runInBand`

## Notes
- No Stripe migration or price-shift scripts were run.
- Runtime env should set `STRIPE_BASIC_PLAN_ID` and `STRIPE_BUILD_PLAN_ID`; `STRIPE_TEST_PLAN_ID` remains a temporary fallback only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,10 +107,13 @@ some tokens on the testnet for making the process simpler.
 
 1. `ENABLE_ACCOUNT_TOPUP`: Enable/disable such functionality (`false` by default)
 2. `FAUCET_URI`: Faucet service API endpoint (Default: `https://faucet-api.cheqd.network/credit`)
-3. `TESTNET_MINIMUM_BALANCE`: Minimum balance on account before it is automatically topped up from the faucet. This value should be expressed as an integer in `CHEQ` tokens, which will then be converted in the background to `ncheq` denomination. Account balance check is carried out on every account creation/login. (Default: 10,000 CHEQ testnet tokens)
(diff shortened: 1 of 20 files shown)


Eengineer1:
Addressed the Copilot review threads in 75392d1:

- Treat `STRIPE_TEST_PLAN_ID` as a Basic alias during the migration window while keeping `STRIPE_BASIC_PLAN_ID` preferred for default assignment.
- Added unit coverage for the legacy Test/Explorer product retaining Testnet/Faucet capability only.
- Reworked account top-up math to use the `ncheq` balance instead of `balances[0]`, and added a safe integer guard for faucet amounts.
- Added missing OpenAPI responses for `404`, `502`, and `503`, and regenerated `src/static/swagger-api.json`.
- Updated migration docs to preserve trialing subscriptions' existing trial end dates.
- Added a defensive default-plan env guard and fixed the Logto comment typo.

Verification:
- `npm run build:app`
- `npm run test:unit -- tests/unit/admin/plan-capabilities.test.ts --runInBand`
- `npm run build:swagger-api`

